PR #13, opened 2014-04-27
Test/QuickCheck/All.hs: fix 'quickCheckAll' to read haskell sources in UTF-8

trofi:
Today I've got an Agda build failure on a box running LANG=C locale (UTF-8
incapable).

```
examples/Heap.hs:151:10:
   Exception when trying to run compile-time code:
     examples/Heap.hs: hGetContents: invalid argument (invalid byte
```

sequence)
         Code: quickCheckAll
       In the expression: $quickCheckAll
       In an equation for `main': main = $quickCheckAll

It's a common error of a 'readFile' assuming locale input. Fixed to read in
UTF-8 in recent GHCs.

Signed-off-by: Sergei Trofimovich slyfox@gentoo.org

### DIFF
--- a/Test/QuickCheck/All.hs
+++ b/Test/QuickCheck/All.hs
@@ -22,6 +22,8 @@ import Data.Char
 import Data.List
 import Control.Monad
 
+import qualified System.IO as S
+
 -- | Test a polymorphic property, defaulting all type variables to 'Integer'.
 --
 -- Invoke as @$('polyQuickCheck' 'prop)@, where @prop@ is a property.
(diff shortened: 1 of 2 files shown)


roberth:
This problem also occurs in `tests/Terminal.hs` https://github.com/nick8325/quickcheck/blob/master/tests/Terminal.hs#L87

```
=== prop_terminal from tests/Terminal.hs:64 ===
*** Failed! Exception: '<file descriptor: 3>: hGetContents: invalid argument (invalid byte sequence)' (after 3 tests and 2 shrinks):
[PutPart "\1050206"]

test-quickcheck-terminal: user error (Pattern match failure in 'do' block at tests/Terminal.hs:94:11-14)
```

PR #34, opened 2014-11-16
Remove unused dependencies from test-suite section

sol:


soenkehahn:
Also it seems that the properties in the `tests` directory are not run during `cabal test`.

soenkehahn:
@nick8325: Is there any reason this is not being merged?

PR #35, opened 2014-11-21
Add shrinkMap and shrinkMapBy

phadej:
I find using this pattern quite often


nick8325:
This looks useful. I don't know why I didn't pull it.

PR #37, opened 2014-12-31
adding new generator combinators: subsetOf, selectOneFrom, shuffled

Javran:
I find these 3 combinators useful, see if there are good enough to be merged into this repo :)

* `subsetOf` generates a subset of the given list.
* `selectOneFrom` is an extension of `elements`: it not only returns the selected element, but also the rest of the choices. this would make sampling without replacement a little bit easier.
* `shuffled` returns a shuffled version of the given list. Additionally, `take n <$> shuffed xs` will then be randomly choosing up to `n` values from `xs` without replacement.

### DIFF
--- a/Test/QuickCheck/Gen.hs
+++ b/Test/QuickCheck/Gen.hs
@@ -20,10 +20,16 @@ import Control.Monad
   ( liftM
   , ap
   , replicateM
+  , foldM
   )
 
 import Control.Applicative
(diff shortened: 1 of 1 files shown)


nick8325:
Thanks! I've merged `shuffled` (renamed to `shuffle`) and `subsetOf` (renamed to `sublistOf` because it doesn't generate a set).

PR #40, opened 2015-01-04
Fix generic shrink, add generic arbitrary and coarbitrary

### LINKED ISSUE #22: genericShrink is probably wrong
Hi,

first many thanks for opening the issue tracker. I think this will benefit QuickCheck a lot.

Now to my issue. Note that it is written with more information than necessary for the QuickCheck maintainer (most of it is probably immediately clear to him) because I wanted to explain the situation to those who haven't spent the last 10 hours on the code.

Currently (as of commit 205f9f9), `genericShrink (0::Int) == [0]`. I don't think this is intended, since the [documentation of `shrink`](http://hackage.haskell.org/package/QuickCheck-2.7.6/docs/Test-QuickCheck.html#v:shrink) mentions that shrinking a value to itself will result in an infinite loop.

The problem is this:

```haskell
instance Typeable a => Subterms (K1 i a) where
  gsubterms (K1 x) =
    case cast x of
      Nothing -> []
      Just y -> [y] -- Bug: this will shrink 0 to [0]
```

I first believed that it can be fixed by `Just y -> shrink y`, but that is not so because then `genericShrink [0::Int] == []`, which is also wrong, it should be `[[]]`.

I noticed that 2 years ago, @bitonic created a `gShrink`/`gArbitrary` implementation (see [here](https://github.com/nh2/quickcheck/commit/146cef9a4375f905af0881684d8c29806b4dbb85]) that doesn't use the `Typeable` constraint, but instead relies on `OverlappingInstances`. In my opinion this is better because it removes a constraint from `genericShrink`, requiring only `Generic`, and the `OverlappingInstances` is only needed in `Arbitrary.hs` to cleanly select the current type and has no effect outside of this file. And of course, it does not come with a run-time penalty while `Typeable` does.

I have created a branch that compares the 5 different ways that I've seen so far - [have a look here](https://github.com/nh2/quickcheck/blob/generic-shrink-evaluation/Test/QuickCheck/Arbitrary.hs#L329):

1. `genericShrinkOrig` - the original `genericShrink` implementation. Wrong because it shrinks 0 to 0.
2. `genericShrinkOrigFixed` - my try to fix it by using `shrink` in the base case. Wrong because it shrinks [0] to [].
3. `genericShrinkBitonic` - same output as `subtermsOrig` (so recursively shrinking subterms is missing) and having the same problem as 1.
4. `genericShrinkBitonicFixed` - my try to fix it by using `shrink` in the base case. Same output as `subtermsOrigFixed` (so recursively shrinking subterms is missing) and having the same problem as 3.
5. `genericShrink`, my improved version, is the apparent solution (please check thorougly if what I say is correct). Explanation:

The core problem is to correctly capture the semantics of `shrink` in the generic instance, that being the [points](http://hackage.haskell.org/package/QuickCheck-2.7.6/docs/Test-QuickCheck.html#v:shrink):

1. Shrink a term to any of its immediate subterms.
2. Recursively apply shrink to all immediate subterms.
3. Type-specific shrinkings such as replacing a constructor by a simpler constructor.

Let's go through them, on the example of `data Tree a = Nil | Branch a (Tree a) (Tree a)`:

1. Implemented by `subterms`. What are the immediate subterms of `Tree a` that are of the same type `Tree a` (since we can only shrink to values of the same type)? The left and the right branch. Not the value `a`, because it is not a tree. This is what the `Typeable` is for: GHC.Generics will give us *all* 3 subterms, `[a, Tree a (left), Tree a (right)]` separated by `*:*`, and we have to `cast` in order to check if each of them really is a Tree. In @bitonic's implementation, this check is done at run-time instead of at compile time by using `class GShrinkBitonic f b` instead of `class SubtermsOrig f` - note how there's one extra type parameter. This allows him to distinguish the two cases `instance Arbitrary a => GShrinkBitonic (K1 i a) a` (subterm is of same type as the term) and `instance GShrinkBitonic (K1 i a) b` (subterm is of different type) using `OvelappingInstances`. Back to the problem: In the `Subterms (K1 i a)` case, the `case x of Just y -> [y]` returns `[0]` when run on `0::Int` because the generic instance of `Int` is basically a `K1` around itself: `from (1::Int) == M1 {unM1 = M1 {unM1 = M1 {unM1 = K1 {unK1 = 1}}}}`.
2. Implemented by `recursivelyShrink`. The idea here is to *keep the same outer structure, but shrink the inner structures*. So basically `Branch x l r` becomes `Branch x (shrink l) (shrink r)`, for any combination of shrinking the left and right part. This is done correctly by the current implementation. For sum types (`RecursivelyShrink (f :+: g)`) just do it inside, for product types (`RecursivelyShrink (f :*: g)`) recursively do it for left or right side *inside* the existing outer structure, for the base case `RecursivelyShrink (K1 i a)` do `grecursivelyShrink (K1 x) = map K1 (shrink x)`, calling `shrink` on the values inside.
3. Clearly, *type-specific shrinkings* are not possible for a generic implementation, since that is for *any* type and thus can't be *type-specific*. Thismeans that shrinking to simpler constructors must always be done manually by the human if some constructors are "simpler" than others (like `Nothing` and `Nil`), and that `genericShrink` can only be used for doing all the non-specific cases, as is explained with `shrink x = shrinkToNil x ++ genericShrink x` in the [QuickCheck docs](http://hackage.haskell.org/package/QuickCheck-2.7.6/docs/Test-QuickCheck.html#v:shrink). For types where the constructors are equally simple, a `shrink = genericShrink` is enough.

So the problem is in 1, namely that `subterms` creates things that are not subterms (`0` isn't a subterm of `0`).

Looking at 1, my `*Fixed` versions are wrong, since I call `shrink` inside of `subterms`, and "Shrink a term to any of its immediate subterms" really means "just take the inner terms of your structure, as they are".

So what's the right `subterms`?

A value can only have true *sub*terms if it has a product type `(:*:)` somewhere.

* If it doesn't, we return `[]`.
* If it does, we return the first value we can grab on either side of the `(:*:)`.

That means we need two type classes for this: An outer `GSubterms` and an inner `GSubtermsIncl`(usive).

* The outer one noticeably returns `gSubterms (K1 _) = []`, and on products calls the inner one: `gSubterms (l :*: r) = gSubtermsIncl l ++ gSubtermsIncl r`.
* The inner (inclusive) one returns everything it can grab, namely `gSubtermsIncl (K1 x) = [x]` if the type matches (this can be done with `GSubtermsIncl (K1 i a) a` or `Typeable`).

As a result, `subterms` returns the first container in a value, unless the value consists only and entirely of that container.

This gives the right result for all `subterms` invocations I've tried so far. Please check if you agree with my reasoning. Again, the working example is `genericShrink` in [here](https://github.com/nh2/quickcheck/blob/generic-shrink-evaluation/Test/QuickCheck/Arbitrary.hs#L329).

I will soon submit a pull request with the cleaned-up version of this.

nh2:
Fixes #22. Much original code contributed by @bitonic. Discussion in #22.

Review +@bitonic

### DIFF
--- a/Test/QuickCheck/Arbitrary.hs
+++ b/Test/QuickCheck/Arbitrary.hs
@@ -232,6 +232,28 @@ instance SumSize (C1 s a) where
   {-# INLINE sumSize #-}
 
 
+-- | This class takes an integer `x` and returns a `gArbitrary` value
+-- for the `x`'th alternative in a sum type.
+class ChooseSum f where
+  chooseSum :: Int -> Gen (f a)
+
+-- Recursive case: Check whether `x` lies in the left or the right side
+-- of the (:+:) split.
+instance ( GArbitrary a, GArbitrary b
+         , SumSize    a, SumSize    b
+         , ChooseSum  a, ChooseSum  b ) => ChooseSum (a :+: b) where
+  chooseSum x = do
+    let sizeL = unTagged2 (sumSize :: Tagged2 a Int)
+    if x <= sizeL
+      then L1 <$> chooseSum x
+      else R1 <$> chooseSum (x - sizeL)
+
+-- Constructor base case.
+instance (GArbitrary a) => ChooseSum (C1 s a) where
+  chooseSum 1 = gArbitrary
+  chooseSum _ = error "chooseSum: BUG"
+
+
 class GArbitrary f where
   gArbitrary :: Gen (f a)
 
@@ -242,7 +264,8 @@ instance (GArbitrary a, GArbitrary b) => GArbitrary (a :*: b) where
   gArbitrary = (:*:) <$> gArbitrary <*> gArbitrary
 
 instance ( GArbitrary a, GArbitrary b
-         , SumSize    a, SumSize    b ) => GArbitrary (a :+: b) where
+         , SumSize    a, SumSize    b
+         , ChooseSum  a, ChooseSum  b ) => GArbitrary (a :+: b) where
   gArbitrary = do
     -- We cannot simply choose with equal probability between the left and
     -- right part of the `a :+: b` (e.g. with `choose (False, True)`),
@@ -252,18 +275,22 @@ instance ( GArbitrary a, GArbitrary b
     -- then a would be chosen just as often as b and c together.
     -- So we first have to compute the number of alternatives using `sumSize`,
     -- and then uniformly sample a number in the corresponding range.
-    -- TODO: The code below could be optimised since the `choose` has to be
-    --       done only once for the whole sum type, not recursively at each
-    --       :+: split. However, this will only give us a 2x speed-up because
-    --       while GHC doesn't guarantee any nesting strategy for :+:, it does
-    --       in practice create binary trees.
-    let sizeL = unTagged2 (sumSize :: Tagged2 a Int)
-        sizeR = unTagged2 (sumSize :: Tagged2 b Int)
-        size = sizeL + sizeR
+    let size = unTagged2 (sumSize :: Tagged2 (a :+: b) Int)
     x <- choose (1, size)
-    if x <= sizeL
-      then L1 <$> gArbitrary
-      else R1 <$> gArbitrary
+    -- Optimisation:
+    -- We could just recursively call `gArbitrary` on the left orright branch
+    -- here, as in
+    --   if x <= sizeL
+    --     then L1 <$> gArbitrary
+    --     else R1 <$> gArbitrary
+    -- but this would unnecessarily sample again in the same sum type, and that
+    -- even though `x` completely determines which alternative to choose,
+    -- and sampling is slow because it needs IO and random numbers.
+    -- So instead we use `chooseSum x` to pick the x'th alternative from the
+    -- current sum type.
+    -- This made it around 50% faster for a sum type with 26 alternatives
+    -- on my computer.
+    chooseSum x
 
 instance GArbitrary a => GArbitrary (M1 i c a) where
   gArbitrary = M1 <$> gArbitrary
--- a/Test/QuickCheck/Arbitrary.hs
+++ b/Test/QuickCheck/Arbitrary.hs
@@ -732,6 +732,19 @@ class CoArbitrary a where
   --   coarbitrary []     = 'variant' 0
   --   coarbitrary (x:xs) = 'variant' 1 . coarbitrary (x,xs)
   -- @
+  --
+  -- In order to generate arbitrary /functions/ of type
+  -- `a1 -> ... -> an -> b`, we must be able to generate arbitrary
+  -- `b`s, and each `ai` given must perturb the `b` created by the
+  -- function. This is what `CoArbitrary` is for.
+  --
+  -- The extension to functions of multiple arguments is done via
+  -- the `instance (Arbitrary a, CoArbitrary b) => CoArbitrary (a -> b)`.
+  --
+  -- The actual generation of arbitrary functions is implemented
+  -- in the `instance (CoArbitrary a, Arbitrary b) => Arbitrary (a -> b)`.
+  --
+  -- For writing properties over functions, see "Test.QuickCheck.Function".
 
   coarbitrary :: a -> Gen b -> Gen b
 
--- a/Test/QuickCheck/Arbitrary.hs
+++ b/Test/QuickCheck/Arbitrary.hs
@@ -3,6 +3,10 @@
 #ifndef NO_GENERICS
 {-# LANGUAGE DefaultSignatures, FlexibleContexts, TypeOperators #-}
 {-# LANGUAGE FlexibleInstances, KindSignatures, ScopedTypeVariables #-}
+{-# LANGUAGE MultiParamTypeClasses, OverlappingInstances #-}
+#endif
+#ifndef NO_SAFE_HASKELL
+{-# LANGUAGE Safe #-}
 #endif
 module Test.QuickCheck.Arbitrary
   (
@@ -20,8 +24,8 @@ module Test.QuickCheck.Arbitrary
   -- ** Helper functions for implementing shrink
 #ifndef NO_GENERICS
   , genericArbitrary   -- :: (Generic a, GArbitrary (Rep a)) => Gen a
-  , genericShrink      -- :: (Generic a, Typeable a, RecursivelyShrink (Rep a), Subterms (Rep a)) => a -> [a]
-  , subterms           -- :: (Generic a, Subterms (Rep a)) => a -> [a]
+  , genericShrink      -- :: (Generic a, Arbitrary a, RecursivelyShrink (Rep a), GSubterms (Rep a) a) => a -> [a]
+  , subterms           -- :: (Generic a, Arbitrary a, GSubterms (Rep a) a) => a -> [a]
   , recursivelyShrink  -- :: (Generic a, RecursivelyShrink (Rep a)) => a -> [a]
 #endif
   , shrinkNothing            -- :: a -> [a]
@@ -104,7 +108,6 @@ import Data.Word(Word, Word8, Word16, Word32, Word64)
 
 #ifndef NO_GENERICS
 import GHC.Generics
-import Data.Typeable
 #endif
 
 --------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


ttuegel:
What is the status of this PR? We're trying to improve the unit test situation in Cabal, and we could really use something like this. We have lots of constructors with many fields, which would be tedious to write Arbitrary instances for, but we already derive Generic.

nh2:
@ttuegel I would also like to know what the status of the PR is, but just as a note: Generic `arbitrary` already works, just not `shrink` and `coarbitrary`.

ttuegel:
@nh2 Do you mean that the Generic `arbitrary` in this PR works, but not the `shrink` or `coarbitrary`? I can't find any reference to a Generic `arbitrary` in `master`.

nick8325:
Hi!

First of all, @nh2, I forgot to thank you and @bitonic for the patch. Thanks!

I'm planning to merge this into the next version of QuickCheck (before GHC 7.10 is out), so there will be generic `arbitrary`, `coarbitrary` and `shrink`. I'm not a big fan of generic `arbitrary`, because you ought to think about test case distribution when writing a property, but the people have spoken and that's what they want :) I may stick a warning in the documentation for `genericArbitrary`, and it won't be the default implementation for `arbitrary` so you will explicitly have to write `arbitrary = genericArbitrary`, but it will be there.

ttuegel:
> I'm not a big fan of generic arbitrary, because you ought to think about test case distribution when writing a property

That's a good point, and a fine argument against making `genericArbitrary` the default. However, I don't think it's always the case that a developer should be thinking about test case distribution. I have been too deliberate about test case distribution before, and the result was test cases that only validated my assumptions about the data my program would be handling. I think the strength of QuickCheck is writing tests that *in*validate your assumptions; for that, a naive distribution is good.

nh2:
> but just as a note: Generic arbitrary already works

@ttuegel Sorry, I got confused. I've had generic `arbitrary` in my Quickcheck for so long that I thought that was already upstreamed.

> @nh2 Do you mean that the Generic arbitrary in this PR works, but not the shrink or coarbitrary? I can't find any reference to a Generic arbitrary in master.

No, you are right. Generic arbitrary is not present in master, but the PR adds it, and it also adds corrected generic shrink and a new generic coarbitrary.

nh2:
@nick8325 

> I'm not a big fan of generic arbitrary, because you ought to think about test case distribution when writing a property

That is a good point, but it's not specific to QuickCheck (e.g. for `Binary` instance might not have the required memory layout) and I do not believe that people who would just write `instance Arbitrary Mytype` will necessarily think more about good generator distribution if they have to write `instance Arbitrary Mytype where arbitrary = genericArbitrary` instead. Instead of sticking the warning on `genericArbitrary`, why not stick it directly on `Arbitrary`?

Also FWIW, without the default signature, GHC 7.10's shortcut syntax `DeriveAnyClass` will not work on `Arbitrary`.

bitonic:
Just wanted to chime in to mention that I would really appreciate for this to be finally merged.

nick8325:
Hi,

I just spoke to Koen, we are a bit worried about `genericArbitrary` still. The problem is the failure mode: if the test case distribution is bad then false properties will pass. I imagine that quite a lot of people will start using `genericArbitrary`, so we have to be very careful to get this right.

As an example, the current `genericArbitrary` will, on the type `[a]`, generate short lists with overwhelming probability (chance of 1/2^n to get a list of length n). This is because it picks `(:)` and `[]` with equal probability. One way to fix this might be to bias the generator towards recursive constructors. But in any case we need to think about what the right test case distribution is, and probably experiment with different ideas.

Koen and I both think that `genericArbitrary` is a useful feature, and we plan to work on getting it right. However, I have to release a new version of QuickCheck before GHC 7.10 comes out, so for the time being I've removed the `genericArbitrary` part of the patch and just merged `genericCoarbitrary` and the fixed `genericShrink`. Sorry! I plan to add `genericArbitrary` once we're confident of the right way to define it, so I'll leave this pull request open for now.

MichaelXavier:
I just absentmindedly implemented this in its own repository and realized that I'm not the first to have done it, hehe. I'm curious if the internal discussion of how to safely implement genericArbitrary got anywhere.

nh2:
> I imagine that quite a lot of people will start using `genericArbitrary`

@nick8325 I believe that when using QuickCheck, you _always_ need to be careful about generator distributions, Generics or not. As I understand it, even if we bias constructors towards recursive constructors, it'll be a rough heuristic, and not something we can guarantee to get right.

I would thus recommend to "solve" this problem through explicit documentation: _"If you use this convenience, this is how it works, take that into account when using it."_

tests/GShrinkExample.hs
```diff
@@ -0,0 +1,17 @@
+{-# LANGUAGE DeriveGeneric, ScopedTypeVariables #-}
+
+module GCoShrinkExample where
```
pthariensflame: I think this should be `GShrinkExample`, not `GCoShrinkExample`, based on the file name.

nh2: Thanks, fixed.

Test/QuickCheck/Arbitrary.hs
```diff
@@ -176,17 +207,118 @@ class Arbitrary a where
...
+-- | `Gen` for generic instances in which each constructor has equal probability
+-- of being chosen.
+genericArbitrary :: (Generic a, GArbitrary (Rep a)) => Gen a
+genericArbitrary = to <$> gArbitrary
+
+
 -- | Shrink a term to any of its immediate subterms,
 -- and also recursively shrink all subterms.
-genericShrink :: (Generic a, Typeable a, RecursivelyShrink (Rep a), Subterms (Rep a)) => a -> [a]
+genericShrink :: (Generic a, Arbitrary a, RecursivelyShrink (Rep a), GSubterms (Rep a) a) => a -> [a]
```
phadej: The `Arbitrary` constraint looks suspicious here, is it really needed?

phadej:
Needs rebase.

MichaelXavier:
Wanted to check on the status of this. This would still be a great thing to have.

oldmanmike:
What's the status of this PR and if there's still work needed, what could I do to help?

MichaelXavier:
Tagging everyone in this thread in, I've gone ahead and fixed the merge conflicts in PR #112 So if everything looks good we can close this and merge that PR, cut a release and enjoy our Generic Arbitrary instances :)

nh2:
That'd be nice.

gbaz:
this package (http://hackage.haskell.org/package/generic-random) and this blog post (https://byorgey.wordpress.com/2016/09/20/the-generic-random-library-part-1-simple-generic-arbitrary-instances/) seem relevant here.

MichaelXavier:
I'll ping this for November. Do we need to fork this library? Does somebody need help to review the patch? I'm accumulating quite a stack of packages that are getting disqualified from stack because of this and I cannot say I understand why this hasn't been merged yet. Please provide some commentary.

RyanGlScott:
For those who are waiting in limbo for #40 to be merged, there's a workaround for the issue of generic `Arbitrary` derivation, as long as you're willing to use `generics-sop` (which requires `base-4.6`/GHC 7.6 at the minimum). Credit goes to @phadej for [this trick](https://github.com/bitemyapp/bloodhound/pull/155):

```haskell
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE ScopedTypeVariables #-}

import           Data.Proxy
import qualified Generics.SOP as SOP
import qualified Generics.SOP.GGP as SOP
import           GHC.Generics
import           Test.QuickCheck

-- Temporary solution for lacking of generic derivation of Arbitrary
-- We use generics-sop, as it's much more concise than directly using GHC.Generics
--
-- This will be unneeded after https://github.com/nick8325/quickcheck/pull/112
-- is merged and released
sopArbitrary :: forall a. (Generic a, SOP.GTo a, SOP.All SOP.SListI (SOP.GCode a), SOP.All2 Arbitrary (SOP.GCode a)) => Gen a
sopArbitrary = fmap SOP.gto sopArbitrary'

sopArbitrary' :: forall xss. (SOP.All SOP.SListI xss, SOP.All2 Arbitrary xss) => Gen (SOP.SOP SOP.I xss)
sopArbitrary' = oneof (map SOP.hsequence $ SOP.apInjs_POP $ SOP.hcpure p arbitrary)
  where
    p :: Proxy Arbitrary
    p = Proxy

-- Example
data Option a = None | Some a
  deriving Generic

instance Arbitrary a => Arbitrary (Option a) where
  arbitrary = sopArbitrary
  shrink = genericShrink
```

nick8325:
Closed (see #112).

PR #41, opened 2015-01-04
In quickCheckAll, when a tyvar uses KindedTV, check the kind

ktvoelker:
It might not really be higher-kinded.

I ran into this on the GHC 7.10 RC, which has a new version of TH (2.10). You *might* be able to call this a bug in TH, and if not a bug, it's certainly an annoying change in behavior.

Also, I am not sure if my patch is general enough. Should a kind like `AppT StarT StarT` be considered plain for QuickCheck's purposes?

The test suite passes with this patch.


ktvoelker:
Oh this is a duplicate. Should've checked the pulls.

PR #49, opened 2015-04-03
More deterministic behavior in regards to exceptions

sol:


nick8325:
Thanks! I'll adapt it to use Test.QuickCheck.Exception instead of Control.Exception.

sol:
Sounds good.

PR #57, opened 2015-07-22
remove unneeded test dependency

soenkehahn:


soenkehahn:
I'm looking into adding a (non-quickcheck) test to the code. Personally I would prefer to use `hspec` for this and since `test-framework` isn't used that may be an option. Is it? Or should I look into `test-framework`?

soenkehahn:
Closing in favor of #34.

PR #60, opened 2015-08-18
Fix a GHC 7.10 warning.

23Skidoo:
`-XOverlappedInstances` is deprecated and is actually unnecessary here, at least on 7.10.


nick8325:
Thanks! I also needed to add some `OVERLAPPING` pragmas, otherwise you get an error when you use `subterms`.

PR #64, opened 2015-08-27
Distinguish between variables and constructors

### LINKED ISSUE #63: Test.QuickCheck.All.monomorphic doesn't work for constructors
`monomorphic` always uses `VarE` internally, which fails if it's given the `Name` of a constructor. For example, to make monomorphic lists:

```
{-# LANGUAGE TemplateHaskell #-}
import Test.QuickCheck.All

nil  = $(Test.QuickCheck.All.monomorphic '[])
cons = $(Test.QuickCheck.All.monomorphic '(:))
```

Loading this in GHCi gives:

```
    Illegal variable name: ‘[]’
    When splicing a TH expression:
      GHC.Types.[] :: [GHC.Integer.Type.Integer]
    In the splice: $(Test.QuickCheck.All.monomorphic '[])
Failed, modules loaded: none.
```

Using `ConE` would allow constructors (after branching on whether the `Name` could be a variable or not).

Warbo:
Fixes #63

Test by loading `tests/Weird.hs` in GHCi; each of the following cases will be run by TemplateHaskell, failing with an error if any isn't right:

 - Monomorphising polymorphic variables, via `$quickCheckAll`. This is the existing behaviour, although it seemed to be untested. I've added `prop_poly` to test this.
 - Monomorphising prefix data constructors. I've added `dummyRun` to test this.
 - Monomorphising infix data constructors, ie. those which begin with `:`. I've added `monoCons` to test this.
 - Monomorphising `[]`, which seems to defy the prefix and infix rules given in the Haskell 2010 Language Report. I've added `monoNil` to test this.

There are no regressions from `cabal test`, although that only seems to run one of the examples.

This was tested with GHC 7.10.2 on NixOS Linux x86.

### DIFF
--- a/Test/QuickCheck/All.hs
+++ b/Test/QuickCheck/All.hs
@@ -66,11 +66,20 @@ monomorphic t = do
   let err msg = error $ msg ++ ": " ++ pprint ty0
   (polys, ctx, ty) <- deconstructType err ty0
   case polys of
-    [] -> return (VarE t)
+    [] -> return (expName t)
     _ -> do
       integer <- [t| Integer |]
       ty' <- monomorphiseType err integer ty
-      return (SigE (VarE t) ty')
+      return (SigE (expName t) ty')
+
+expName :: Name -> Exp
+expName n = if isVar n then VarE n else ConE n
+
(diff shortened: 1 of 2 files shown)


nick8325:
Thanks!

PR #66, opened 2015-09-28
Weaken (Ord a) constraint on Arbitrary (NonZero a)

adamse:
Weaken Ord to Eq, the instance did not make use of the Ord constraint.


nick8325:
Thanks!

PR #68, opened 2015-10-21
Add build flag to disable TFGen

erratic-pattern:
I'd like to make a tool that makes it possible to build tests that compare function outputs from multiple Haskell backends against each other, in order to ensure consistent behavior on multiple backends. Currently I can't use Arbitrary for this tool, because I need to ensure that the randomized output is the same on all backends, Adding a build flag to disable TFGen should make this possible, albeit somewhat hacky since I have to somehow ensure that QuickCheck was always built with this flag. Perhaps it would be better to instead make it possible to choose exactly which generator to use? This could be done, for example, by using a SomeGen existential, but that would require the use of language extensions. Is there another way to allow this?

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -124,7 +124,7 @@ library
   if impl (ghc < 7.4)
     cpp-options: -DNO_SAFE_HASKELL
 
-  -- Use tf-random on newer GHCs. Build with noTDRandom flag to disable.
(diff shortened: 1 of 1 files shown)


MaximilianAlgehed:
Closing this as stale. Feel free to open an issue and clarify the problem this is trying to solve.

PR #69, opened 2015-11-24
Generic derivation of Arbitrary via DefaultSignature

dmjio:
This should naïvely allow us to generically derive instances of `Arbitrary`. cc @fizruk

```haskell
{-# LANGAUGE DeriveGeneric #-}
module Main where

import GHC.Generics
import Test.QuickCheck

data Person = Person {
   name :: String
 , age :: Int
} deriving (Generic, Show)

instance Arbitrary Person

{-- derived instance looks like
instance Arbitrary Person where
  arbitrary = Person <$> arbitrary <*> arbitrary
--}

main :: IO ()
main = print =<< ((generate arbitrary) :: IO Person)

### DIFF
--- a/Test/QuickCheck/Arbitrary.hs
+++ b/Test/QuickCheck/Arbitrary.hs
@@ -125,6 +125,10 @@ import GHC.Generics
 class Arbitrary a where
   -- | A generator for values of the given type.
   arbitrary :: Gen a
+#ifndef NO_GENERICS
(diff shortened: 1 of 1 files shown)


dmjio:
This is a duplicate of https://github.com/nick8325/quickcheck/pull/40, verbatim implementation it seems. cc: @nh2... except @nh2 also included Generic coarbitrary.

PR #70, opened 2015-12-21
Make compatible with template-haskell 2.11 / GHC 8.0

hvr:
While at it, this also makes the QuickCheck package convenient for
unregisterised and/or cross-compilers lacking TH support with latest
Cabal solvers by declaring the need for Template-Haskell in
`other-extensions` as a dependency, which then directs the cabal solver
toggle off the TH support flag if that extension is not advertised by
GHC.

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -50,6 +50,7 @@ flag base4point8
 
 flag templateHaskell
   Description: Build Test.QuickCheck.All, which uses Template Haskell.
+  Default: True
 
 library
   -- Choose which library versions to use.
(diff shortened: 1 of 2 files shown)


thomie:
Please merge. It works.

nick8325:
Thanks, merged! I think this was already addressed by commit 0625361078776b9e9b1f0a389b3bc67fcd348de1 but the `other-extensions` thing is nice.

thomie:
>  I think this was already addressed by commit 0625361

Maybe, but `cabal install QuickCheck` is still failing with GHC 8, so if 1cd45e8 (this patch) could also be released to hackage, that would be great.

nick8325:
QuickCheck 2.8.2 is out now which builds on GHC 8.

thomie:
Great! Thank you for maintaining QuickCheck.

PR #72, opened 2016-01-21
ZipList and other instances

phadej:


nick8325:
Thanks, this looks good!

PR #76, opened 2016-02-09
Use nats dependency with older compilers / base

phadej:
Includes commit from https://github.com/nick8325/quickcheck/pull/75

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -100,8 +100,12 @@ library
   else
       cpp-options: -DNO_NATURALS
 
-  if impl(ghc < 8.0)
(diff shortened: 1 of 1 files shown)


nick8325:
I'm going to leave this open for now, see my comment on #78.

phadej:
@nick8325 I see. I added semigroups in this PR, as you probably gonna accept them together if you do.

I'd be very grateful for consistency here, as I got bitten in e.g. https://github.com/bos/aeson/pull/363

nick8325:
It would seem a bit silly to only provide these instances for GHC 8. I suppose it's fine really to add dependencies on packages that have since been merged into `base`.

Issue #74, opened 2016-02-09
Drop support for GHC <7


phadej:
I tried to make a matrix with GHC 6.12.3, and it starts to be quite troublesome to compile. Maybe one could get further by disabling tests, but then `tested-with` would lie.

Particularly, seems that I broke `binary-0.8` for GHC <7.0 (because it isn't tested)

*EDIT* this would highly simplify cabal file too.

hvr:
Unless there's actually somebody using or testing latest QuickCheck with GHCs prior to 7.0, I second this. 

GHC 7.0 (which was released in 2010) is most likely the oldest GHC you can expect on Linux distribution which are not yet EOL'ed (with "enterprise" distros being the ones we need to worry about most).

see also https://github.com/haskell/cabal/issues/3108#issuecomment-180064508

phadej:
Thanks. I'll update my other pull request during today.

PR #78, opened 2016-02-10
NonEmpty instances

phadej:
Conflicts with https://github.com/nick8325/quickcheck/pull/76 as this introduces new flag which should be taken there into account too.

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -97,6 +97,14 @@ library
   if impl(ghc < 7.10)
     cpp-options: -DNO_NATURALS
 
+  
+  -- Semigroups, particularly NonEmpty. Appeared in base-4.9 / GHC 8.0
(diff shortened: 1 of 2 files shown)


phadej:
It seems that it hard to make old `cabal-install`s to construct build plans automatically, of course flags could be given manually.

Yet, by dropping GHC<7 support https://github.com/nick8325/quickcheck/issues/74 and using `impl(ghc >= 7.10)` conditions would make solver's job trivial.

nick8325:
I pulled your change but without the dependency on `semigroups`, i.e., it only adds the instances for GHC 8.

I'm a bit reluctant to add more dependencies to QuickCheck, but maybe in this case (and your `nats` ticket) it's appropriate for consistency. Anyway I will have a think about it and maybe add the dependency later.

PR #79, opened 2016-02-11
Drop support for ghc < 7

phadej:
Resolves https://github.com/nick8325/quickcheck/issues/74


phadej:
ping @nick8325 . GHC 8.0 build should succeed after caches are wiped (GHC was updated in between)

nick8325:
Thanks!

PR #83, opened 2016-02-22
Write specialized instance for sequences

treeowl:
Replace list-based `Arbitrary` instance with one using
primitives from `Data.Sequence`. This avoids the inefficiency
of `fromList` in general, and takes advantage of fast splitting
and appending.

Note: there may well be a way to improve shrinking further;
this commit simply translates the version for lists into
the language of sequences.

### DIFF
--- a/Test/QuickCheck/Arbitrary.hs
+++ b/Test/QuickCheck/Arbitrary.hs
@@ -532,8 +532,28 @@ instance Arbitrary a => Arbitrary (IntMap.IntMap a) where
   arbitrary = fmap IntMap.fromList arbitrary
   shrink = map IntMap.fromList . shrink . IntMap.toList
 instance Arbitrary a => Arbitrary (Sequence.Seq a) where
-  arbitrary = fmap Sequence.fromList arbitrary
-  shrink = map Sequence.fromList . shrink . toList
+  arbitrary = seqOf arbitrary
+  shrink xs = shrinkSeq shrink xs
+
+-- | Shrink a list of values given a shrinking function for individual values.
+shrinkSeq :: (a -> [a]) -> Sequence.Seq a -> [Sequence.Seq a]
+shrinkSeq shr xs = concat [ removes k n xs | k <- takeWhile (>0) (iterate (`div`2) n) ]
+                 ++ shrinkOne xs
+ where
+  n = Sequence.length xs
+
+  shrinkOne s =
+    case Sequence.viewl s of
+      Sequence.EmptyL -> []
+      x Sequence.:< xs -> [x Sequence.<| xs | x' <- shr x ]
+                  ++ [ x Sequence.<| xs' | xs' <- shrinkOne xs ]
+
+  removes k n xs
+    | k > n     = []
+    | Sequence.null xs2  = [Sequence.empty]
+    | otherwise = xs2 : map (xs1 Sequence.><) (removes k (n-k) xs2)
+   where
+    (xs1,xs2) = Sequence.splitAt k xs
 
 -- Arbitrary instance for Ziplist
 instance Arbitrary a => Arbitrary (ZipList a) where
--- a/Test/QuickCheck/Gen.hs
+++ b/Test/QuickCheck/Gen.hs
@@ -35,6 +35,7 @@ import Control.Arrow
 import Test.QuickCheck.Random
 import Data.List
 import Data.Ord
+import qualified Data.Sequence as Seq
 
 --------------------------------------------------------------------------
 -- ** Generator type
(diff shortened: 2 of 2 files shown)


treeowl:
The failure is clearly not my fault.

nick8325:
Have you been having performance problems with generation of `Seq`s? I would be interested to hear if so.

It seems that your new implementation is a bit slower than the old one. At least it is when I run the following benchmarks:

```haskell
seqGen1, seqGen2 :: Gen (Seq Int)
seqGen1 = seqOf arbitrary
seqGen2 = fmap fromList arbitrary

main =
  defaultMain [
    bench "1" (nfIO (generate (vectorOf 100 seqGen1))),
    bench "2" (nfIO (generate (vectorOf 100 seqGen2))) ]
```

Here are the results I get, showing the existing `Arbitrary` instance to be slightly faster:

```
benchmarking 1
time                 1.545 ms   (1.533 ms .. 1.554 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 1.564 ms   (1.560 ms .. 1.576 ms)
std dev              20.29 μs   (8.966 μs .. 39.34 μs)

benchmarking 2
time                 1.409 ms   (1.388 ms .. 1.436 ms)
                     0.998 R²   (0.996 R² .. 1.000 R²)
mean                 1.386 ms   (1.380 ms .. 1.401 ms)
std dev              28.20 μs   (14.06 μs .. 57.86 μs)
```

Not sure why this is - but it does make me doubt that it's worth making the `Arbitrary` instance more complex in search of performance gains.

I'm also sceptical that the change to `shrink` will improve performance - I'd expect the bottleneck in shrinking to be running the test cases rather than generating the shrink candidates.

treeowl:
That is surprising indeed, especially considering the *apparently* balanced nature of the `Monad` implementation of `Gen`. I wonder if it would make any difference to replace

```haskell
  MkGen m >>= k =
    MkGen (\r n ->
      let (r1,r2)  = split r
          MkGen m' = k (m r1 n)
       in m' r2 n
    )
```
with
```haskell
  MkGen m >>= k =
    MkGen (\r n -> case split r of
      (r1,r2) -> unGen (k (m r1 n)) r2 n)
```

which at least *appears* nicer in general, and may or may not play better with `replicateA`. Unfortunately, I'm not really set up properly for benchmarking at the moment.

As for why I'm even thinking about this: `fromList` in `Data.Sequence` is kind of a beast, and it's really hard for me to imagine that there isn't *some* faster way to generate random sequences. You may well be right about the shrinks; on the other hand, it may also be possible to generate *more* shrinks in some more "natural" fashion for sequences. I'll need to think that through some more.

nick8325:
I just tried replacing the `let` with a `case`. It does make a difference - the current `Seq` generator gets slower! Your generator stays the same speed.

It could be that some subcomputations are not generating any random numbers and thus the `case` forces `split`s that are never used.

I'm surprised to see, though, that generating a plain list is only slightly slower than generating a `Seq` (and much slower than I would expect). So maybe there is a performance problem somewhere. I'll try and look into it.

treeowl:
OK, I think I may have tracked down why my implementation is slower: the `Applicative` instance has an extra `split` because `ap` applies `>>=` twice:

```haskell
fs `ap` xs = fs >>= \f -> xs >>= \x -> return (f x)
```

The `return (f x)` obviously doesn't need a generator, but it gets one anyway. The obvious fix there is to change the `Applicative` instance to

```haskell
MkGen m1 <*> MkGen m2 = MkGen $ \r n ->
    let (r1, r2) = split r
    in m1 r1 n (m2 r2 n)
```

I'm pretty confident that change is strictly an improvement in all cases, and it should improve my sequence instance substantially.

nick8325:
Sounds like a very sensible suggestion - but I'm afraid it makes your generator *slower*!

I have no idea why that could be - something funny must be going on.

treeowl:
I'm flabbergasted. That's starting to sound like something crazy with the inliner or something. I'm going to submit a pull request for just that `Applicative` change and some similar things, and then let's see how to adjust everything else.

treeowl:
Ooh, I actually had another interesting idea, which can reasonably be expected to give dramatically better performance when using `TFRandom`. Just wait for this next pull request :-)

treeowl:
OK, I won't keep it hidden. `TFRandom` offers `splitn`, which splits a seed into up to 2^32 new seeds in one go, turning it into a function `Word32 -> TFGen`. So to generate a list/sequence, once we've figured out how *long* the list/sequence should be, we can build the result in 2^32-element chunks and potentially save a *lot* of time. This will be especially zippy using recent versions of `Data.Sequence`, for which `mapWithIndex` is very efficient.

nick8325:
Good idea! QuickCheck's `coarbitrary` already uses `splitn`, so the best way to do this is to generate say `f :: Int -> a` and then return the list `map f [0..whatever]`.

I tried this just now and unfortunately it didn't make much difference. So I suspect the bottleneck with generation is somewhere else. The generated core doesn't look great, some things aren't specialised, some things aren't fused, etc. - I suspect that's where we need to look for performance gains.

treeowl:
For *long* sequences, either the `splitn` approach with `fromFunction` or the current approach using `Data.Sequence.replicateA` should have much better laziness (hence GC and cache) properties than `replicateM`. The benchmark you showed only looked at tiny ones, if I read it right, where something else dominates. As for the general generated core appearance, I'm betting the haphazard introduction of laziness via lazy pattern bindings is partially responsible. That will wreak all sorts of havoc in the optimizer. Laziness is a glorious thing, but it takes a light touch. You may well be right about specialization too&mdash;I'll try to dig into it properly this week and see if I can work something out.

nick8325:
Closing this as it made things slower.

PR #85, opened 2016-02-23
Improve Applicative and Monad instances

treeowl:
Avoid splitting the generator twice for `<*>`.

Avoid splitting the generator in the unlikely cases
of `*>`, `<*`, and `>>`.

Add specialized versions of liftA3, liftA4, and liftA5
that use TFRandom splitn.

Use `Functor` and `Applicative` instances (much) more.


nick8325:
I'm concerned that this patch adds quite a bit of code to QuickCheck and complicates the generators. It doesn't look much fun to maintain.

I think we should improve the definition of `<*>` and replace `liftM` with `fmap` and leave the rest for now. Don't forget that with `ApplicativeDo`, GHC can do many of these transformations itself in theory.

Having said that, GHC won't e.g. translate `liftM2` into `liftA2` unless `Control.Monad` is compiled with `ApplicativeDo`. But it seems like, if splitting is a performance bottleneck, instead of this patch we should:
* Ask the GHC people to compile `Control.Monad` with `ApplicativeDo`.
* Think if there is a way to make `Gen`'s `Applicative` instance use `splitn` automatically.

*But* I don't think either step is necessary unless we establish that there is a solid performance gain to be had here.

PR #86, opened 2016-02-23
Make QCGen split like underlying distribution

treeowl:
We can use `GeneralizeNewtypeDeriving` to use the underlying
`split` precisely. Otherwise, we can at least be as strict
as it is. I strongly believe we should be adding laziness only
strategically, rather than all over the place as at present.


nick8325:
Thanks, I merged this change but removed the `GeneralizedNewtypeDeriving` stuff (I believe just using `case` has the same semantics).

PR #89, opened 2016-02-29
Add Aribtrary instances up to 10-tuples

Shimuuar:
I needed Arbitrary instances for lare tuples so I've added them

Only thing that worries me is shrink. Cross product of all possible shrinks for all tuple elements for say 10-tuple could become LARGE


nick8325:
Thanks a lot! Shrinking isn't a problem because shrinking for tuples doesn't do the cross product, it shrinks one element at a time.

PR #91, opened 2016-03-02
Don't depend on Template Haskell or tf-random when installed using Haste.

valderman:
Haste is offended by C dependencies and Template Haskell, so let's shut them off when building with Haste.


nick8325:
Thanks!

Issue #99, opened 2016-07-01
Please specify package version bounds for dependencies


mrkkrp:
Quick Check 2.8.2 does not work with *any* `template-haskell`. In particular, it does not work with version 2.11. Please take time and specify package version bounds.

phadej:
Could you elaborate? What doesn't work? http://matrix.hackage.haskell.org/package/QuickCheck and travis say `QuickCheck` is green.

mrkkrp:
https://travis-ci.org/stackbuilders/stache/jobs/141392203#L326

nick8325:
OK, it seems that Cabal is installing a newer version of `template-haskell` than the one shipped with GHC?!

I had no idea that was possible, I thought `template-haskell` was a wired-in package.

In that case the CPP macros in `Test/QuickCheck/All.hs` are wrong. They check the version of GHC and they need  to check the `template-haskell` version instead. I will push a patch ASAP.

PR #100, opened 2016-07-01
Use correct in TH code

phadej:
Resolves https://github.com/nick8325/quickcheck/issues/99


nick8325:
Aha, fantastic! Thanks!

PR #103, opened 2016-07-10
Add "again" to "forAll"

Icelandjack:
Haven't tested this but in the latest version this only executes once

```haskell
>>> verboseCheck $ forAll (arbitrarySizedBoundedIntegral @Int) (\x -> x /= minBound)
Passed:  
1
+++ OK, passed 1 tests. 
```

I have to use `again`, seems it was forgotten

```haskell
>>> quickCheck $ again $ forAll (arbitrarySizedBoundedIntegral @Int) (\x -> x /= minBound)
+++ OK, passed 100 tests.
```


nick8325:
Oh no! Thanks for pointing this out. I fixed it by defining `forAll` in terms of `forAllShrink` (which is how I thought it was already...)

PR #104, opened 2016-07-12
Export fun

phadej:
- Export `Fun` type, `apply`, `Function(..)`, and `functionMap` from `Test.QuickCheck`
    - I often find I need just `Fun` and `apply` when writing test-suites. It could be very convenient if they are re-exported from `Test.QuickCheck`.
    - I intentionally didn't re-export `Fun` constructor
   -  For this I made `Test.QuickCheck.Function` and `Test.QuickCheck.Poly` *Safe*
- Also added documentation to 'Fun' and 'apply', so `Test.QuickCheck` haddock have something about them.
    - Added a type to 'Fn', for some reason I was getting `(t -> t) -> Fun t t` with GHC7.10.3 (in GHC-8.0 it seems to be fixed)
    - You can see this issue in old haddocks: http://hackage.haskell.org/package/QuickCheck-2.8.2/docs/Test-QuickCheck-Function.html#v:Fn

### DIFF
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -176,10 +176,10 @@ module Test.QuickCheck
 
     -- ** Functions
   , Fun
+  , appFun
 #if defined(__GLASGOW_HASKELL__) && __GLASGOW_HASKELL__ >= 708
   , pattern Fn
 #endif
-  , apply
   , Function (..)
   , functionMap
 
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -69,6 +69,9 @@ showable functions used for testing higher-order functions and
 #ifndef NO_SAFE_HASKELL
 {-# LANGUAGE Safe #-}
 #endif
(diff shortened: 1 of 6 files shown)


nick8325:
Hi! Sorry for being so slow.

This change looks good. Would you mind rebasing it to the latest HEAD, and removing the version bump from `QuickCheck.cabal`?

Your patch has some other changes as well (marking `Test.QuickCheck.{Poly,Monadic}` as `Safe`, not using `GeneralizedNewtypeDeriving` in `Test.QuickCheck.Poly`). Is that deliberate? I can merge it too if you like.

phadej:
@nick8325 I don't remember anywhere, but somewhere I wanted `Test.QuickCheck.Monadic` to be *Safe* module. I guess it doesn't harm, and shouldn't pose problems later

phadej:
@nick8325  rebased.

nick8325:
Thanks!

PR #105, opened 2016-07-28
Add a `chooseAny` Gen combinator

mikemckibben:
I often find myself needing to generate a random value over the entire natural range of an instance of `Random a`. Additionally, some instances of `Random` ignore the range argument of `randomR` and using the `choose` combinator doesn't make sense for these types, e.g. Data.UUID. This pull request adds a `chooseAny` combinator defined as

```haskell
chooseAny :: Random a => Gen a
```


nick8325:
Thanks!

PR #108, opened 2016-08-05
Add MonadGen class

### LINKED ISSUE #107: Add MTL-style class for Gen
I found myself wanting this today.

```haskell
class Monad m => MonadGen m where
  liftGen :: Gen a -> m a

instance MonadGen Gen where
  liftGen = id

instance MonadGen m => MonadGen (StateT s m) where
  liftGen = lift . liftGen

-- and so on
```

treeowl:
* Add an mtl-style `MonadGen` class for monads int which one can
lift `Gen` values.

* Add the obvious instances. ~~I did not have the time to consider
whether there are sensible instances available for `ContT` or
even the version of `ListT` in `transformers`, so I left those
out.~~ I don't know if the `transformers` `ListT` makes sense for this,
so I left it out.

Fixes #107

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -51,6 +51,7 @@ library
     Test.QuickCheck,
     Test.QuickCheck.Arbitrary,
     Test.QuickCheck.Gen,
+    Test.QuickCheck.Gen.Class,
     Test.QuickCheck.Gen.Unsafe,
     Test.QuickCheck.Monadic,
     Test.QuickCheck.Modifiers,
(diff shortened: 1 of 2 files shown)


treeowl:
Ed Kmett says the simple lifting is correct for `ContT`.

nick8325:
Thanks! Looks useful.

nick8325:
I recently bumped into the `QuickCheck-GenT` package: http://hackage.haskell.org/package/QuickCheck-GenT. This provides the `MonadGen` class but in a more complete way (it also lifts the various combinators such as `listOf` to `MonadGen`).

I would rather not do that automatic lifting in the main `Test.QuickCheck` module (because of the potential breakage to users), so on second thoughts maybe we should point people at `QuickCheck-GenT` instead of having our own version.

PR #112, opened 2016-09-01
Fix generic shrink rebase

MichaelXavier:
I've gone ahead and taken pull #40 and merged master into it. Everything compiles and tests pass on my machine (it would be a good idea to set up travis and I'd be happy to do that if you wanted, or I can even do that in a separate PR). Please let me know if there's anything else that needs to be done before this goes out. The thread [here](https://github.com/fpco/stackage/issues/1734) lit a fire under me to get this fixed soon. On a lot of my projects for lack of a generic arbitrary I've been using the `derive` library which uses TH to the same effect. That library apparently is causing problems on stackage, so I'd like to switch over to using Generic instead.


MichaelXavier:
Scratch any mention of Travis. I can clearly see that this project has Travis enabled, not sure what I was thinking there.

Test/QuickCheck/Arbitrary.hs
```diff
@@ -27,6 +27,7 @@ module Test.QuickCheck.Arbitrary
   , arbitraryBoundedEnum          -- :: (Bounded a, Enum a) => Gen a
   -- ** Helper functions for implementing shrink
 #ifndef NO_GENERICS
+  , genericArbitrary   -- :: (Generic a, GArbitrary (Rep a)) => Gen a
```
RyanGlScott: Perhaps this should re-exported from `Test.QuickCheck`, since the functions below it are re-exported as well?

RyanGlScott:
I haven't had a chance to look at this until recently, but this looks really nice! I do have a couple of comments (one of which is inline).

One thing: GHC 8.0 [introduced representation types](http://hackage.haskell.org/package/base-4.9.0.0/docs/GHC-Generics.html#t:UAddr) for certain unlifted data types. For example, `UInt` represents `Int#`, and you could have an `GArbitrary` instance for it. Something like:

```haskell
instance GArbitrary UInt where
  arbitrary = do
    I# i# <- arbitrary
    return $ UInt i#
```

Similarly for `UAddr`, `UChar`, `UDouble`, `UFloat`, and `UWord`. Although these were introduced in GHC 8.0, it's possibly to backport support for them using the `generic-deriving` library (and this would also allow the generic machinery in `QuickCheck` to support GHC 7.0 as well!)

If this sounds reasonable, I could open a PR on your branch that incorporates these changes.

MichaelXavier:
I hadn't even heard of that. No objection here, it looks like generic-deriving is present in stackage which is what motivated this whole thing in the first place (being able to drop the `deriving` dependency in some libraries).

RyanGlScott:
@MichaelXavier: I've opened https://github.com/Soostone/quickcheck/pull/1, which implements my suggestions above.

MichaelXavier:
Merged and tests have passed.

bitemyapp:
@RyanGlScott is this ready to go forward now?

MichaelXavier:
@bitemyapp Your comment popped up right as I loaded this page to prod this issue :laughing: I think maybe you meant that for @nick8325? @RyanGlScott's PR has been merged into this one and the tests all pass.

MichaelXavier:
@RyanGlScott @nick8325 Bumping this again as its been about a month since last action on this. What is blocking this moving forward?

RyanGlScott:
(BTW, it should be noted that I am _not_ a maintainer of `QuickCheck`, just an enthusiastic user. Several people have tried pinging me with requests to merge things, but I don't wield that kind of power :) )

MichaelXavier:
@RyanGlScott whoops. Sorry for the misunderstanding!

joshvera:
@nick8325 Would you mind reviewing this? I think a lot of users would find this feature useful.

RyanGlScott:
Hey @nick8325,

It's been almost two months since this was opened, and I'd like to make a plug for this again, and also give some reasons as to why a lot of people really want this feature (and preferably sooner than later!).

At the moment, there are a [large number of packages](https://github.com/fpco/stackage/issues/1734#issuecomment-243494548) that have been using the `derive` package to quickly generate `Arbitrary` instances for their large ADTs (as implementing those `Arbitrary` instances by hand can be tedious). Unfortunately, `haskell-src-exts` (an important dependency of `derive`) recently released a new major version, and the maintainer of `derive` isn't keeping it up-to-date anymore, so users of `derive` are looking for alternatives.

This PR would be a perfect answer to this problem. The `GHC.Generics`-based code that is provides would make creating new `Arbitrary` instances cheap and cheerful:

```haskell
data LargeADT = ...
  deriving Generic

instance Arbitrary Generic
```

That's it! Now libraries can shed their bulky `derive` dependency and just use `QuickCheck` directly.

I know personally that I have several libraries that I'm holding back from working on due to not having this feature, and I think [others are in a similar boat](https://github.com/fpco/stackage/issues/1734#issuecomment-244141184).

Would you mind looking at this PR sometime?

Test/QuickCheck/Arbitrary.hs
```diff
@@ -140,10 +155,35 @@ import Data.Functor.Constant
...
+-- | Random generation and shrinking of values.
+#endif
 class Arbitrary a where
   -- | A generator for values of the given type.
   arbitrary :: Gen a
+#if !defined(NO_GENERICS) && __GLASGOW_HASKELL__ >= 702
+  default arbitrary :: (Generic a, GArbitrary (Rep a)) => Gen a
+  arbitrary = genericArbitrary
+#else
+  arbitrary = error "no default generator"
```
phadej: should this just be left off?

Test/QuickCheck/Arbitrary.hs
```diff
@@ -1,15 +1,23 @@
...
+#if __GLASGOW_HASKELL__ >= 702
+{-# LANGUAGE DefaultSignatures #-}
+#endif
 #if __GLASGOW_HASKELL__ < 710
 {-# LANGUAGE OverlappingInstances  #-}
 #endif
 #endif
 #ifndef NO_SAFE_HASKELL
-{-# LANGUAGE Safe #-}
+{-# LANGUAGE Trustworthy #-}
```
phadej: This is breaking change, btw. What causes this not being Safe anymore?

RyanGlScott: Sadly, in GHC 7.2, `GHC.Generics` was inferred as `Unsafe`, so importing it destroys safety. You could probably add some more CPP guards to get it to be safe on GHC 7.4 and later, if you so desired.

nick8325:
Sorry for taking such a very long time to reply.

First, I would like to point people at the [`testing-feat`](http://hackage.haskell.org/package/testing-feat) package, which can generate `Arbitrary` instances automatically, and works hard to generate good-quality test data. Given a datatype `A`, you can write:

```
deriveEnumerable ''A
instance Arbitrary A where arbitrary = sized uniform
```

This gives you an `Arbitrary` instance where the test data is chosen uniformly at random from all values of the appropriate size (size being the number of constructors in the generated value). This is not always exactly what you want, but for algebraic datatypes it normally gives a good variety of test cases.

---

Unfortunately there are some big problems with the patch in this pull request.

Firstly, it doesn't look at the size parameter at all, which means that for recursive datatypes it generates huge or even infinite values. For example, if I define a type of ternary trees...

```
> data Tree = Leaf Int | Node Tree Tree Tree
>   deriving (Show, Generic, Arbitrary)
```

...then it commonly (about a third of the time) produces an infinite tree:

```
> sample (arbitrary :: Gen Tree)
Node (Leaf 0) (Leaf 0) (Node (Node (Node (Node (Leaf 0) (Node (Leaf 0) (Node (Node (Node (Node (Node (Leaf 0) (Leaf 0) (Node (Leaf 0) (Node (Node (Leaf 0) (Node (Node (Node (Node (Leaf 0) (Node (Leaf 0) (Leaf 0) (Leaf 0)) (Node (Leaf 0) (Node (Node (Node (Node (Leaf 0) (Leaf 0) (Node (Node (Leaf 0) (Node (Leaf 0) (Node (Node (Node (Leaf 0) (Node (Node (Node (Leaf 0) (Leaf 0) (Leaf 0)) (Leaf 0) (Leaf 0)) (Leaf 0) (Node (Leaf 0) (Node (Node (Leaf 0) (Node (Node (Leaf 0) (Leaf 0) (Leaf 0)) (Leaf 0) (Leaf 0)) (Leaf 0)).......
```

Another problem is if you have a list-like structure:

```
data List = Nil | Cons Int List
  deriving (Show, Generic, Arbitrary)
```

Because the generator has a 50% chance of choosing `Nil` at each point, the probability that we generate a list of length `n` decreases exponentially with `n`, so we tend to get extremely short lists:

```
len :: List -> Int
len Nil = 0
len (Cons _ xs) = 1+len xs

> quickCheck (\xs -> collect (len xs) True)
+++ OK, passed 100 tests:
52% 0
25% 1
15% 2
 6% 3
 1% 6
 1% 5
```

We generated 100 random lists, and only 2 of them had more than 3 elements. If you test your program using this particular generator, QuickCheck is unlikely to find any but the most trivial bugs.

If the test data distribution is bad, like it is above, it's a disaster because it _seems_ like everything is fine. QuickCheck will tell you that 100 tests passed - it won't tell you that the test data was rubbish. In other words, QuickCheck will report that your code is thoroughly tested while missing many of the bugs, which is about the worst thing it can do. As far as I'm concerned, a generic `arbitrary` MUST produce reasonable test data no matter what you throw at it; if it fails to do so, it silently breaks your testing.

Unfortunately, I don't know how to solve this. While we could fix the specific problems above with this patch, there are other datatypes where it won't work reliably and I don't think those can be fixed (I can expand on this if wanted). So I doubt that we will ever have generic `arbitrary` in QuickCheck, but if you want a generator for free you can get it from `testing-feat`. Having said that, you will get more mileage out of your testing if you think carefully about test case distribution, so a thoughtful, handwritten generator is still best whenever possible.

nh2:
@nick8325 I'm not convinced by your argument, because:

A badly written, naive generic `Arbitrary` instance is just as bad as a badly written, naive by-hand instance.

The point of using `Generic` (at least for many of us) is not to generate instances where doing so is hard. Its benefit is that we can avoid manual instances where writing generic instances is easy and correct.

In most types in real-world programs are simple non-recursive sum or product types, where a generic instance would do all fine. There's a real-world productivity benefit from having it.

If you have a data type, like `List`, for which the generic instance would create a bad distribution and your tests less effective, then it is *wrong code* and against the QuickCheck documentation to use the generic instance for it. Just like it is wrong code to use a generic `Binary` instance for something that will be badly serialised by it, or a generic `ToJSON` instance that doesn't generate the syntax you need.

nh2:
@nick8325 Also, if you care about people not accidentally writing bad instances, I'm sure there's a good compromise in making such instances explicit.

For example, how about (in slightly exaggerated terminology):

```
data LargeADT = ...
  deriving Generic

instance TrivialDumbArbitrary LargeADT

instance Arbitrary LargeADT where arbitrary = trivialDumbArbitrary
```

This would be only one line of boilerplate for a user, and make it very explicit that an instance is used that does not try to handle corner cases like the well-known one you presented; the `arbitrary = trivialDumbArbitrary` could be easily commented on with something like `-- safe to use this here because LargeADT is just a flat sum type`.

What do you think?

RyanGlScott:
I agree with everything that @nh2 said.

`testing-feat` just won't work for my use cases. It doesn't support the same range of GHC versions that `QuickCheck` does, and it requires Template Haskell, which is a no-go for several of my projects.

I don't really care whether or not `genericArbitrary` is the default implementation of `arbitrary`. In fact, I'd be perfectly fine with making `genericArbitrary` a separate function that's marked with __`WARNING: THIS MIGHT PRODUCE A BAD DISTRIBUTION IF YOU AREN'T CAREFUL`__ in bold text. I just want to have the ability to quickly mock up `Arbitrary` instances for non-trivial ADTs without having to type out the soul-crushing amount of boilerplate that is currently required.

Even if the algorithm behind `genericArbitrary` isn't ideal, it's much, much better than the current status quo. Having `genericArbitrary` in `QuickCheck` would provide a central location where it can be maintained (and improved!) instead of me having to copy and paste an [enormous chunk of code](https://github.com/RyanGlScott/text-show/blob/8e92bd9f45f156e4219037e7ff886527935d96d1/tests/Instances/Utils/GenericArbitrary.hs) in every single project of mine that requires this.

Lysxia:
Searching for [`"arbitrary arbitrary"` on Github](https://github.com/search?utf8=%E2%9C%93&q=language%3Ahaskell+%22arbitrary+arbitrary%22&type=Code) can give an idea of how useful such a simple `genericArbitrary` would be, despite its imperfection. (This search string mostly finds large products, and surely there are plenty large sums as well that this doesn't find.)

There are  of newcomers

PR #114, opened 2016-09-02
Testable (Gen prop) should be always tried again

### LINKED ISSUE #113: Number of QuickCheck test cases is inconsistent
@jtdaugherty reports that somewhere between QuickCheck  2.8.2 and 2.9.1 a bug (or a silent and undocumented in the Changelog change in semantics) was introduced: https://github.com/feuerbach/tasty/issues/147

Could you look into it?

phadej:
Fix #113


nick8325:
Thanks! I think this is the right approach.

PR #116, opened 2016-09-09
Changed listOf to take the size parameter better into account

NorfairKing:
This ensures that the magnitude of the `size` parameter is arbitrarily distributed over the elements of the list that is generated.
Otherwise a long list cannot contain small elements.

### DIFF
--- a/Test/QuickCheck/Gen.hs
+++ b/Test/QuickCheck/Gen.hs
@@ -190,15 +190,25 @@ growingElements xs = sized $ \n -> elements (take (1 `max` (n * k `div` 100)) xs
 -- size parameter.
 listOf :: Gen a -> Gen [a]
 listOf gen = sized $ \n ->
-  do k <- choose (0,n)
-     vectorOf k gen
+  do size <- choose (0,n)
+     pars <- arbPartition size
+     mapM (\i -> resize i gen) pars
 
 -- | Generates a non-empty list of random length. The maximum length
 -- depends on the size parameter.
 listOf1 :: Gen a -> Gen [a]
 listOf1 gen = sized $ \n ->
-  do k <- choose (1,1 `max` n)
-     vectorOf k gen
+  do size <- choose (1,1 `max` n)
+     pars <- arbPartition size
+     mapM (\i -> resize i gen) pars
+
+arbPartition :: Int -> Gen [Int]
+arbPartition k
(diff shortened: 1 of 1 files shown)


NorfairKing:
To be clear:

```
quickCheck $ forAll (resize 100 $ listOf $ sized $ \n -> pure $ n + 1) $ all (> 100)
+++ OK, passed 100 tests.
```

This is a serious problem.

nick8325:
The size parameter represents the *maximum* size of value to generate. For example, `listOf` itself chooses a number between 0 and the current size as the length of the list. As long as your generators interpret the size parameter as maximum size there is absolutely no problem. For example, trying `sample (resize 100 (arbitrary :: Gen [Int]))` will show lists with a variety of elements. Note that `sized $ \n -> pure (n+1)` is not a sensible generator for this reason.

There are two problems with changing `listOf` to work as you suggest:

* We would get much shorter lists. Currently a list of size 100 has an expected length of 50.
* By choosing the size of each partition greedily, we get a biased distribution - larger elements are concentrated towards the beginning of the list.

NorfairKing:
'The maximum size value to generate' representation does not hold for the current `listOf` generator.
The current `listOf` generator generates values of maximum size `size * size` (maximum size `size` for each element of a list of maximum length `size`).

Old version:
```
> (length . concat . concat . concat . concat) <$> generate (resize 25 $ listOf $ listOf $ listOf $ listOf $ listOf (arbitrary :: Gen Int))
257361
```

New version:

```
> (length . concat . concat . concat . concat) <$> generate (resize 25 $ listOf $ listOf $ listOf $ listOf $ listOf (arbitrary :: Gen Int))
3
```


Feel free to change the way partitions are chosen if that is the problem.

nick8325:
Right, QuickCheck has a problem with deeply-nested lists - they get exponentially longer as the nesting increases. With your idea they get exponentially _shorter_ instead - I'm not sure it's an improvement.

I suppose it's more accurate to say that the size represents the maximum size of the spine of that particular structure.

NorfairKing:
Even if this patch does not get merged, can we agree that this is a problem?

*If the size parameter does not get distributed there is a quadratic blowup in memory usage.*

I'd like the `size` parameter to be an indication of the total memory usage of the generated data.

PR #119, opened 2016-10-08
Add newtype modifiers for Arbitrary String/Char instances that take into account Unicode

berdario:


Test/QuickCheck/Modifiers.hs
```diff
@@ -343,4 +355,100 @@ instance (Arbitrary a, ShrinkState s a) => Arbitrary (Shrinking s a) where
...
+
+asciiGen :: Gen Char
+asciiGen = choose ('\0', '\127')
+
+instance Arbitrary ASCII where
+  arbitrary = ASCII `fmap` asciiGen
+
+--------------------------------------------------------------------------
+-- | @Latin1@: latin1 Char, same Arbitrary as Char itself.
+newtype Latin1 = Latin1 {getLatin1 :: Char}
```
phadej: For me it's surprising that `Char` instance generates only `Latin1` characters. Could you add a haddock to `Arbitrary Char` saying it?

Also how the `-----` top bar affects the haddocks?

berdario: Yeah, ideally I'd change the `Arbitrary Char` instance to return unicode Chars... in the meanwhile I'll add an haddock

not sure how `-----` affects it, but it was already present for every other newtype

Test/QuickCheck/Arbitrary.hs
```diff
@@ -579,6 +579,7 @@ instance Arbitrary Word64 where
   shrink    = shrinkIntegral
 
 instance Arbitrary Char where
+  -- | This currently only generates Latin1 Chars, if you want Unicode use the @Unicode@ newtype
```
phadej: It probably should be above the instance. Could you generate docs locally to check?

Test/QuickCheck/Modifiers.hs
nick8325: I think it's just a historical accident that `Arbitrary Char` only generates Latin1. We should just bite the bullet and change it to Unicode.

Is a Latin1 generator useful in its own right? Or would it be enough with Unicode and (7-bit) ASCII?

berdario: Yeah, I'm also :+1: for switching to Unicode right away. If this will cause breakage it's because people have latent Unicode bugs in their codebase for which they're not aware, and getting tests to fail will be useful.

I added the Latin1 newtype only so that people would have an easy migration path if they want to retain the existing logic and don't want to fix their code

I don't think it adds much value over the ASCII newtype, but you never know on which behavior downstream code might be relying upon

btw, one small difference is the shrink: I think that the set of unicode codepoints is too big to meaningfully explore the space when shrinking. The same might not be true for the ASCII newtype, but for simplicity I didn't bother replicating Latin1's logic into it

Test/QuickCheck/Modifiers.hs
```diff
@@ -343,4 +355,100 @@ instance (Arbitrary a, ShrinkState s a) => Arbitrary (Shrinking s a) where
...
+
+--------------------------------------------------------------------------
+-- | @Unicode@: yields unicode Char.
+newtype Unicode = Unicode {getUnicode :: Char}
+  deriving ( Eq, Ord, Show, Read )
+
+unicodeGen' :: Gen Char
+unicodeGen' = oneof [choose (minBound, '\55296'), -- Skip high and low surrogates
+                     choose ('\57344', maxBound)]
+
```
berdario: I just realized... distribution of values here might be skewed...
should I rather go with `suchThat` instead of `oneof`?

phadej: you can use [`frequency`](http://hackage.haskell.org/package/QuickCheck-2.9.2/docs/Test-QuickCheck.html#v:frequency)

berdario: Yup, but the numbers would be awkward... I'll calculate them when I get back to my laptop

berdario: Actually, they're coprimes, so it'd need to be something like

```
frequency [(55297, choose (minBound, '\55296')) , -- Skip high and low surrogates
                   (1056768,  choose ('\57344', maxBound))]

```

nick8325:
I moved the generators themselves into `Test.QuickCheck.Arbitrary`, and removed Latin1 and `AllUnicode` (we can add them if there's demand for it).

I also removed the modifiers that generate single characters, so we only have the ones that generate whole strings. If someone wants to generate a single character they can use the generators directly.

berdario:
Thank you

Sorry for not reacting earlier. I'd have applied myself the changes you suggested, but Iife intervened

Test/QuickCheck/Modifiers.hs
berdario: AllUnicode could be useful for code that handles surrogate pairs (and needs to detect mismatched pairs)...

`Text.pack`ing an invalid surrogate pair will just yield `U+FFFD`, so it's indeed not very useful for most common code. And I expect that code deailing with UTF-16 will most often deal directly with `Word`s, rather than `Char`s

nick8325:
No problem, and thanks for the patch!

PR #132, opened 2016-12-08
Add more Arbitrary1 instances

phadej:
I left out tuples and monoid/semigroup (and some other) wrappers.

Also changed the `NonEmpty` shrink a bit, now it's better (no filtering).

### DIFF
--- a/Test/QuickCheck/Arbitrary.hs
+++ b/Test/QuickCheck/Arbitrary.hs
@@ -166,6 +166,9 @@ import qualified Data.Monoid as Monoid
 #ifndef NO_TRANSFORMERS
 import Data.Functor.Identity
 import Data.Functor.Constant
+import Data.Functor.Product
+import Data.Functor.Sum
+import Data.Functor.Compose
 #endif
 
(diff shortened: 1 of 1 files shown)


Test/QuickCheck/Arbitrary.hs
```diff
@@ -432,9 +435,15 @@ instance Arbitrary a => Arbitrary [a] where
   shrink = shrink1
 
 #ifndef NO_NONEMPTY
+instance Arbitrary1 NonEmpty where
+  liftArbitrary arb = liftM2 (:|) arb (liftArbitrary arb)
+  liftShrink shr (x :| xs) =
+      [ x' :| xs | x' <- shr x ] ++
+      [ x :| xs' | xs' <- liftShrink shr xs ]
+
```
nick8325: This is a bit worse than before because it won't try removing the first element from the list. We should probably just keep the old shrinking approach (it adds very little overhead).

phadej: Good point, I'll revert

Issue #130, opened 2016-12-05
Different approach to shrinking


Icelandjack:
May be of interest: [Integrated vs type based shrinking](http://hypothesis.works/articles/integrated-shrinking/)

reiddraper:
I [proposed this](https://mail.haskell.org/pipermail/libraries/2013-November/021674.html) several years ago, and unfortunately have heard nothing from the QuickCheck maintainers.

phadej:
I wonder, whether `RoseTree` approach will work with `Fun`?

thumphries:
I use an [implementation](https://github.com/ambiata/disorder.hs/tree/master/disorder-jack/) of this idea every day, it really improves the ergonomics of property testing! Would love to see it in QuickCheck.

nick8325:
It's a nice approach which I have a lot of time for, but it also has a couple of drawbacks.

1. This style of `Gen` works well as an `Applicative`, but less well as a `Monad`. For example, if you generate pairs by writing `(,) <$> arbitrary <*> arbitrary` then both components can be shrunk independently. However, if you write the following generator instead:

```
do
  x <- arbitrary
  y <- arbitrary
  return (x, y)
```

then the shrinking behaviour is: first `x` is shrunk, then `y`, but once it's finished shrinking `y` it _won't_ try shrinking `x` again. (It wouldn't be safe to do so because if the generator made decisions based on `x`'s value, you might end up in an infinite shrinking loop.)

`ApplicativeDo` helps here of course but it still seems a bit fragile.

2. It's a bit less flexible, because your shrinking strategy has to follow the recursive structure of your generator.

For example, QuickCheck's shrinking for lists (function `shrinkList`) first tries to remove half the list, then a quarter, then an eighth, and so on until it gets down to single elements. This is just defined using normal list functions. It's harder to do this with shrinking built into the generators - I think you'd have to write your generator in a divide-and-conquer style, where you recursively generate both halves of the list, then append them together while trying to shrink them both to the empty list.

Now imagine that we want to change shrinking so that it (say) also tries to remove the middle third of the list. This is a simple change in QuickCheck at the moment, but I don't know how to add it in this other design. You might try adding it as an extra shrinking step at the end, but that means that once you remove the middle third of a list you can no longer shrink the individual elements. (I suppose that this is related to the problem in number 1 above - shrinking becomes sensitive to the order in which you generate things, which is not really the case in current QuickCheck.)

[...]

PR #136, opened 2016-12-14
Rose

phadej:
Related to https://github.com/nick8325/quickcheck/issues/130

FWIW, I'm not sure whether this should be in QuickCheck right now, probably better to make `Rose` an own library, and wait it to stabilize before merging into main package.

### DIFF
--- a/Test/QuickCheck/Rose.hs
+++ b/Test/QuickCheck/Rose.hs
@@ -1,0 +1,74 @@
+-- | A Rose 'Tree' based combinators to write generators and shrinkers
+-- simultaenously.
+--
+-- >>> let reven = fmap (*2) rarbitrary
+-- >>> quickCheck $ rForAll reven $ \x -> x < 4
+-- *** Failed! Falsifiable (after 6 tests and 1 shrink):     
+-- 4
+--
+-- See <http://hypothesis.works/articles/integrated-shrinking/>
+module Test.QuickCheck.Rose
+  (
+  -- * Generators
+    RGen (..)
+  , rarbitrary
+  , roneof
+  , rsized
+  , rscale
+  -- * Properties
+  , rForAll
+  )
+ where
+
+import Control.Applicative
+import Control.Monad (join)
+import Data.Traversable (traverse)
+import Data.Tree (Tree (..))
+import Test.QuickCheck
+import Test.QuickCheck.Property (Property (..))
+
+-- | A generator for values of type @a@ with built-in shrinking.
+--
+-- /Note:/ you should use applicative combinators (or @ApplicativeDo@),
+-- as much as possible to get quality shrinking.
+newtype RGen a = RGen { runRGen :: Gen (Tree a) }
+
+instance Functor RGen where
+    fmap f (RGen g) = RGen (fmap (fmap f) g)
+
+instance Applicative RGen where
+    pure = RGen . pure . pure
+    RGen f <*> RGen x = RGen $ (<*>) <$> f <*> x
+
+instance Monad RGen where
+    return = pure
+    RGen g >>= f = RGen $ do
+        tree  <- g
+        tree' <- traverse (runRGen . f) tree
+        pure $ join tree'
+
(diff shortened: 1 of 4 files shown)


phadej:
The problem with `oneof`: it doesn't know how to try previous "generator" stuff before. Have to think about that.

Test/QuickCheck/Rose.hs
```diff
@@ -0,0 +1,97 @@
...
+instance Applicative RGen where
+    pure = RGen . pure . pure
+    RGen f <*> RGen x = RGen $ (<*>) <$> f <*> x
+
+instance Monad RGen where
+    return = pure
+    RGen g >>= f = RGen $ do
+        tree  <- g
+        tree' <- traverse (runRGen . f) tree
+        pure $ join tree'
```
jystic: I don't think this is the bind implementation you want.

`traverse` will split the random number generator for every node in `tree`, and this will cause unpredictable shrinking, because every time we shrink the `tree` it causes the rest of the computation `f` to be executed with a wildly different seed.

See the PR on `disorder-jack` where I fixed this same problem: https://github.com/ambiata/disorder.hs/pull/99

jystic:
> I'm not sure whether this should be in QuickCheck right now, probably better to make `Rose` an own library

Fwiw, I plan to split `disorder-jack` out in to its own repository and release it on Hackage sometime soon (probably January)

Test/QuickCheck/Rose.hs
phadej: I see.

phadej:
@jystic split would be great, it looks like you have given `disorder-jack` some thought already, but I'd like to comment few things right away:

- `containers` has `Tree`, does it lack some functionality, which forced you to duplicate the type?
- `arbitraryJack :: Arbitrary a => Jack a` would be great
- have you tried something like `oneof'` (with a prime) in this PR, for my example it somehow improved the shrink results
- have you given thought how your `recursive` could be written, so the shrink results contains the subterms. In this PR, I had  to write shrink for `Ty` manually, to include subterms, which btw would been there by default with `genericShrink` of `QuickCheck` (didn't wanted to use generics in this example though)

jystic:
> containers has Tree, does it lack some functionality, which forced you to duplicate the type?

The bind for `Tree` in `containers` walks the tree in the opposite direction, this results in shrinks which are not minimal, see the [comment](https://github.com/ambiata/disorder.hs/blob/master/disorder-jack/src/Disorder/Jack/Tree.hs#L42-L44)

> arbitraryJack :: Arbitrary a => Jack a would be great

See [arbitrary](https://github.com/ambiata/disorder.hs/blob/master/disorder-jack/src/Disorder/Jack/Combinators.hs#L389-L392)

> have you tried something like oneof' (with a prime) in this PR, for my example it somehow improved the shrink results

Yes, [oneOf](https://github.com/ambiata/disorder.hs/blob/master/disorder-jack/src/Disorder/Jack/Combinators.hs#L195-L203) in Jack already works that way, but it just uses `choose`, which itself shrinks to `0` and thereby selects earlier elements in the list if possible.

> have you given thought how your recursive could be written, so the shrink results contains the subterms. In this PR, I had to write shrink for Ty manually, to include subterms, which btw would been there by default with genericShrink of QuickCheck (didn't wanted to use generics in this example though)

Jack has a combinator [reshrink](https://github.com/ambiata/disorder.hs/blob/master/disorder-jack/src/Disorder/Jack/Core.hs#L93-L96) that lets you apply additional shrinks in addition to the existing shrink tree. 

There is a [test](https://github.com/ambiata/disorder.hs/blob/master/disorder-jack/test/Test/Disorder/Jack/Minimal.hs) which demonstrates how it's used to achieve a minimal shrink in situations where you need shrink results to contain subterms.

phadej:
@jystic great!

phadej:
I'll give `disorder-jack` a try when it's released!

moodmosaic:
FWIW, `disorder-jack` is the ancestor of [Hedgehog](http://hackage.haskell.org/package/hedgehog).

PR #138, opened 2016-12-18
Parallel generators / implementation matching

paolino:
I am testing 2 implementations for the same API one against the other one. Without (=!=)  I need some more boilerplate interleaved in the generating code.

```
test2Algo :: (Arbitrary a, Eq a) => Algo c a -> Algo d a -> Gen Bool
test2Algo x y = do
    (n :: Int) <- arbitrary
    (q,r,ts) <- (\f -> foldM f (newQueue x,newQueue y,[]) [0..n-1]) $ 
        \(q,r,ts) _ -> do
            xs <- listOf1 arbitrary
            k <- choose (0, length xs - 1) 
            let t x q = unloadn x k <$> elements 
                        [meld x (create x xs) q, flip (load x) xs q]
            (q',es) <- t x q
            (r',es') <- t y r
            return (q', r' , (es == es') : ts)
    return $ unload x q == unload y r && all id ts

test :: Gen Bool
test = test2Algo (naive :: Algo [] Int) efficient
```

With (=!=) I can just generate the ([a],[[a]]) and match them after so the testing function stay cleaner

```
genAlgo :: Arbitrary a => Algo c a -> Gen ([a], [[a]])
genAlgo x = do
    (n :: Int) <- arbitrary
    (q,ts) <- (\f -> foldM f (newQueue x,[]) [0..n-1]) $ 
        \(q,ts) _ -> do
            xs <- listOf1 arbitrary
            k <- choose (0, length xs - 1) 
            let t x q = unloadn x k <$> elements 
                        [meld x (create x xs) q, flip (load x) xs q]
            (q',es) <- t x q
            return (q', es : ts)
    return $ (unload x q, ts)

test :: Gen Bool
test = genAlgo (naive :: Algo [] Int)  =!= genAlgo efficient
```

The full code of the example is [here](https://gist.github.com/paolino/793ca314b5ff0876f99bec59fc091fae)

### DIFF
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -122,6 +122,7 @@ module Test.QuickCheck
   , infiniteList
     -- ** Running a generator
   , generate
+  , replayer
     -- ** Generator debugging
   , sample
   , sample'
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -116,6 +116,8 @@ module Test.QuickCheck
   , infiniteListOf
   , shuffle
   , sublistOf
+  , parallel
+  , (=!=)
     -- ** Generators which use Arbitrary
   , vector
   , orderedList
--- a/Test/QuickCheck/Gen.hs
+++ b/Test/QuickCheck/Gen.hs
@@ -102,6 +102,13 @@ generate (MkGen g) =
   do r <- newQCGen
      return (g r 30)
 
+-- | create a reproduceable generating environment
+replayer :: IO (Gen a -> a)
+replayer = do
+    r <- newQCGen
+    let repeater (MkGen g) = g r 30
+    return $ repeater
+
 -- | Generates some example values.
 sample' :: Gen a -> IO [a]
 sample' g =
--- a/Test/QuickCheck/Gen.hs
+++ b/Test/QuickCheck/Gen.hs
@@ -102,12 +102,21 @@ generate (MkGen g) =
   do r <- newQCGen
      return (g r 30)
 
--- | create a reproduceable generating environment
+-- | a reproduceable generate
 replayer :: IO (Gen a -> a)
 replayer = do
     r <- newQCGen
     let repeater (MkGen g) = g r 30
     return $ repeater
+
+-- | evaluate 2 generators with the same seed
+parallel :: Gen a -> Gen a -> Gen (a,a)
+parallel (MkGen x) (MkGen y) = MkGen $ \g n -> 
+    (x g n, y g n)
+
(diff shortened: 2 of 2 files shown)


nick8325:
It's better to divide your property into a part which generates test cases and a part which runs them. Here is some code I came up with, where the test case is represented as a list of "steps":

```haskell
data Step a = MeldCreate Int [a] | Load Int [a] deriving Show
instance Arbitrary a => Arbitrary (Step a) where
  arbitrary = oneof [pair MeldCreate, pair Load]
    where
      pair f = do
        xs <- listOf1 arbitrary
        k  <- choose (0, length xs - 1)
        return (f k xs)

interpret :: Algo c a -> [Step a] -> ([a], [[a]])
interpret alg xs = (unload alg q, ts)
  where
    (q, ts) = foldl' step (newQueue alg, []) xs
    step (q, ts) x =
      let
        (q', es) =
          case x of
            MeldCreate k xs ->
              unloadn alg k $ meld alg (create alg xs) q
            Load k xs ->
              unloadn alg k $ load alg q xs
      in (q', es:ts)

prop_same_behaviour :: [Step Int] -> Property
prop_same_behaviour xs =
  interpret naive xs === interpret efficient xs

main = quickCheck prop_same_behaviour
```

Why is this better? Well, one reason is that if the property fails you get a counterexample, which you don't with the `=!=` approach. Another reason is that using `=!=` is very fragile. For example, if `genAlgo` looked at the contents of `q` when deciding what function to generate next, like in the following code, it wouldn't work:

```haskell
-- This will not work with =!=
genAlgo x = do
   ... $ \(q, ts) -> do
      if isEmpty x q then do
         ... generate something ...
      else
         ... generate something else ...
```

The reason this doesn't work is because, if the two queue implementations differ, you can end up generating two different test cases (in one implementation `isEmpty` can return true, and in the other it can return false, and then the rest of the test case will be different). So this is a very delicate way to do testing, I wouldn't recommend it; better to separate generating the tests from running them.

By the way, `replayer` is very similar to an existing QuickCheck combinator, the function `delay :: Gen (Gen a -> a)` in the module `Test.QuickCheck.Gen.Unsafe`. In particular, you can define `replayer = generate delay`, so it's not really needed. The reason `delay` (and by extension `=!=`) are unsafe is because they make `Gen` violate the monad laws: for example the property `x =!= x` will be true, but `x =!= (return () >> x)` will be false, because they use the random number seed in different ways. This makes `x /= (return () >> x)` in the presence of `=!=`. Any combinator which allows you to re-use a random number seed has this problem. Such combinators are a very specialised tool, not something to be used in normal testing (`delay` is used for example in implementing the `Arbitrary (a -> b)` instance).

PR #140, opened 2016-12-24
Fixing LICENSE to be clearly a BSD-3-Clause

pegartillo95:
As you have selected the BSD-3_Clauses on Github I understand that is your purpose but writting in there "All rights reserved" at the same time is quite confusing.


nick8325:
Thanks! It seems that "All rights reserved" was required in the distant past but is no longer needed.

PR #142, opened 2016-12-26
Add genSize :: Gen Int

chris-martin:
This is a trivial little definition, but it helps write slightly nicer-looking generators.

As the docs in this PR demonstrate, one can rewrite

```haskell
listOf gen = sized $ \n -> do
  k <- choose (0, n)
  vectorOf k gen
```

as

```haskell
listOf gen = do
  n <- genSize
  k <- choose (0, n)
  vectorOf k gen
```

The latter style makes more sense to me, and I think it's easier to teach, so it'd be nice to have it in the library :)

### DIFF
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -106,6 +106,7 @@ module Test.QuickCheck
   , elements
   , growingElements
   , sized
+  , genSize
   , resize
   , scale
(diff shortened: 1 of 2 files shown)


nick8325:
@koengit, what do you think?

nick8325:
Thanks!

PR #147, opened 2017-01-07
Add Args.maxShrinks

phadej:
cc @thumphries

### DIFF
--- a/Test/QuickCheck/Test.hs
+++ b/Test/QuickCheck/Test.hs
@@ -60,6 +60,9 @@ data Args
     -- ^ Size to use for the biggest test cases
   , chatty          :: Bool
     -- ^ Whether to print anything
+  , maxShrinks      :: Int
+    -- ^ Maximum number of shrinks to before giving up. Setting this to zero
+    --   turns shrinking off.
   }
  deriving ( Show, Read )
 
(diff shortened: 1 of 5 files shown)


Test/QuickCheck/Test.hs
```diff
@@ -371,6 +375,10 @@ foundFailure st res ts =
 localMin :: State -> P.Result -> P.Result -> [Rose P.Result] -> IO (Int, Int, Int)
 localMin st MkResult{P.theException = Just e} lastRes _
   | isInterrupt e = localMinFound st lastRes
+-- Don't try to shrink for too long
+localMin st res _ ts
+  | numSuccessShrinks st + numTotTryShrinks st >= numTotMaxShrinks st =
```
thumphries: Could `numSuccessShrinks st + numTotTryShrinks st` overflow?

thumphries:
Thanks for implementing this!

👍

Test/QuickCheck/Test.hs
phadej: I don't think so, as the check is done on every iteration.

nick8325:
Using `Test.QuickCheck.Function` can result in a very large number of shrink steps (sometimes into the thousands), and if you don't run shrinking to completion you get a function which is not showable. So I would rather have the default be unlimited shrinking, and the user can explicitly set `maxShrinks` if they want less.

nick8325:
P.S. thanks for the patch!

phadej:
amended

nick8325:
Huh, Travis fails now. Looks like `prop_B1` needs to be updated? I think instead of `$quickCheckAll` you want (IIRC) `$forAllProperties (quickCheckWithResult stdArgs { maxShrinks = 10000 })`.

phadej:
amended once again

PR #149, opened 2017-01-13
Don't use {-# OVERLAPPING #-} pragmas on GHC < 7.10

RyanGlScott:
A minor thing: when you build the current version of `QuickCheck` on GHC < 7.10, it emits warnings. Example [here](https://travis-ci.org/sol/doctest/jobs/191673800#L806):

```
Test/QuickCheck/Arbitrary.hs:317:10: Warning: Unrecognised pragma

Test/QuickCheck/Arbitrary.hs:320:10: Warning: Unrecognised pragma
```

That's because GHC didn't support the `{-# OVERLAPPING #-}` pragma until GHC 7.10 (before then, it just used the `-XOverlappingInstances` extension. To avoid this, I used CPP to make `OVERLAPPING_` expand to `{-# OVERLAPPING #-}` on recent GHCs, but expand to nothing on older GHCs. This is the same trick that [`aeson` uses.](https://github.com/bos/aeson/blob/d470ac61874b4dc80b88a44a0c2d967430a8d5ce/include/overlapping-compat.h)

### DIFF
--- a/Test/QuickCheck/Arbitrary.hs
+++ b/Test/QuickCheck/Arbitrary.hs
@@ -5,8 +5,11 @@
 {-# LANGUAGE DefaultSignatures, FlexibleContexts, TypeOperators #-}
 {-# LANGUAGE FlexibleInstances, KindSignatures, ScopedTypeVariables #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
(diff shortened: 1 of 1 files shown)


nick8325:
Thanks!

Issue #127, opened 2016-11-19
Opportunity for Cont/Codensity


Icelandjack:
Just brainstorming, nesting [`forAll`](https://hackage.haskell.org/package/QuickCheck-2.9.2/docs/Test-QuickCheck.html#v:forAll), [`shrinking`](https://hackage.haskell.org/package/QuickCheck-2.9.2/docs/Test-QuickCheck.html#v:shrinking) isn't very pretty and they return a continuation (and the type of [`sized`](https://hackage.haskell.org/package/QuickCheck-2.9.2/docs/Test-QuickCheck.html#v:sized) is a [`Codensity`](https://hackage.haskell.org/package/kan-extensions-5.0.1/docs/Control-Monad-Codensity.html#t:Codensity) of `Gen`)

```haskell
forAll       @_ @Property :: Show a => Gen a               -> (a -> Property) -> Property
forAllShrink @_ @Property :: Show a => Gen a -> (a -> [a]) -> (a -> Property) -> Property
shrinking       @Property :: (a -> [a]) -> a               -> (a -> Property) -> Property

sized :: (Int -> Gen a) -> Gen a
```

This might be useful (mostly the `do-notation`)

```haskell
(·) = fmap.fmap 

cont . forAll       :: Show a => Gen a               -> Cont Property a
cont · forAllShrink :: Show a => Gen a -> (a -> [a]) -> Cont Property a
cont · shrinking    :: (a -> [a]) -> a               -> Cont Property a

newtype Codensity m a = Codensity (forall xx. (a -> m xx) -> m xx)

Codensity sized :: Codensity Gen Int
```

Thoughts?

nick8325:
This is very much like what `Test.QuickCheck.Monadic` does, isn't it? I mean, your `cont . forAll` is the same as `pick`, I think.

Perhaps it would make sense to polish `Test.QuickCheck.Monadic` to make it easier to use for this purpose? For example, we could add some `Testable` instances for `PropertyM`.

Icelandjack:
That might be fun, the second one probably doesn't make sense

```haskell
instance IO ~ m => Testable (PropertyM m a) where
  property :: PropertyM IO a -> Property
  property = monadicIO

instance Gen ~ m => Testable (PropertyM m a) where
  property :: PropertyM Gen a -> Property
  property prop = property $ do
    propGen <- monadic' prop
    fmap property propGen
```

Icelandjack:
[...]

PR #161, opened 2017-04-14
Update PropertyM Applicative/Monad instance

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

https://github.com/nick8325/quickcheck/issues/127

### DIFF
--- a/Test/QuickCheck/Monadic.hs
+++ b/Test/QuickCheck/Monadic.hs
@@ -81,6 +81,9 @@ import Test.QuickCheck.Gen.Unsafe
 import Test.QuickCheck.Property
 
 import Control.Monad(liftM, liftM2)
+#if MIN_VERSION_base(4,9,0)
+import Control.Monad.Fail(MonadFail, fail)
+#endif
 
 import Control.Monad.ST
 import Control.Applicative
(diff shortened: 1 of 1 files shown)


Test/QuickCheck/Monadic.hs
```diff
@@ -105,14 +108,23 @@ newtype PropertyM m a =
...
-instance Monad m => Applicative (PropertyM m) where
-  pure = return
-  (<*>) = liftM2 ($)
+instance Applicative (PropertyM m) where
+  pure x  = MkPropertyM ($ x)
+  f <*> v = MkPropertyM $ \ c -> unPropertyM f $ \ g -> unPropertyM v (c . g)
+
+#if MIN_VERSION_base(4,9,0)
+instance Monad (PropertyM m) where
+  return x = MkPropertyM ($ x)
```
phadej: just `return = pure` would be clearer? Or actually omitting it fully in base-4.9 case?

gliptak: @Icelandjack Would you comment? Thanks

nick8325:
Thanks for the patch! There is a bit of duplicated code here, which I would like to avoid. In particular:

* `<*>` and `>>=` share a lot of the same code. To avoid this, we could define `bind :: PropertyM m a -> (a -> PropertyM m b) -> PropertyM m b` and use that to define `<*>` and `>>=`.
* The `Monad` instance for `PropertyM` appears twice (with slightly different code). I would like to keep the existing definitions as they work fine. Instead of duplicating it we can use CPP to a) remove the `Monad m` superclass constraint, and b) add the line `instance Monad m => MonadFail (PropertyM m) where` before the definition of `fail`.
* The duplication between `return` and `pure` that @phadej pointed out.

Also, please could you not use `MIN_VERSION_base` but make the cabal file define a preprocessor macro, e.g. `NO_MONAD_FAIL`, on base < 4.9. This makes it possible to load QuickCheck directly into GHCi which is handy for development/debugging.

gliptak:
@nick8325 Thank you for helping out the refactoring

PR #162, opened 2017-04-20
Add missing Function Word instance

Bodigrim:
There are instances of `Function` for `Word8/16/32/64`, but not for `Word` itself.


nick8325:
Thanks!

Issue #158, opened 2017-03-20
Decouple Arbitrary from QuickCheck


kvanberendonck:
So that I can ship libraries with Arbitrary instances without adding QuickCheck (and hence all its dependencies) as a library dependency.

RyanGlScott:
_All_ its dependencies? You make it sound like there's an enormous amount of them, when really `QuickCheck`'s dependencies are quite svelte by Haskell standards. With GHC 8.0, you only need `primitive`, `random`, and `tf-random`, which is quite reasonble IMHO.

kvanberendonck:
It just seems a little unreasonable (and this is opinionated) to have 'QuickCheck' as a dependency for a library if you want to offer 'QuickCheck' support for a couple of users who might find it useful in their unit testing.

kvanberendonck:
If there was a standalone lightweight `arbitrary` package, you could just merge all the *-instances packages directly into their actual packages. Additionally, it works around a problem with stack/cabal where the dependency system means you can't have tests that depend on benchmarks or vice-verca and if you want to decouple your arbitrary instances from your unit tests without pulling in QuickCheck as a dependency in your main app/lib whilst still allowing maximum access you have to make a separate utility library (otherwise circular reference).

RyanGlScott:
...but depending on slightly modified version of `QuickCheck` that only exports `Arbitrary` _is_ reasonable? I'm not trying to be snarky here, I genuinely don't understand the reasoning. I think including `QuickCheck` as a library dependency is quite reasonable.

kvanberendonck:
OK maybe my suggestion was unreasonable. Do you mind leaving it open to gather some more data?

kvanberendonck:
The way I imagined this would work is making a new library `Data.Arbitrary` and having `QuickCheck.Arbitrary` simply re-export from there so that new versions of `QuickCheck` allow people to import `Data.Arbitrary` for other purposes.

sol:
I would like to see this too.  Ideally even going one step further and splitting `QuickCheck` into two packages e.g. `QuickCheck-core` and `QuickCheck`.  `QuickCheck-core` would provide everything we need to

1. Run QuickCheck properties
2. Extend QuickCheck

with minimal dependencies.

`QuickCheck` on the other hand would depend on `QuickCheck-core` and provide all the additional instances that we currently have.

thumphries:
[...]

PR #166, opened 2017-05-19
Remove light flag; migrate instances to quickcheck-instances

phadej:
Related to #158. See https://github.com/phadej/qc-instances/pull/2 for corresponding PR in `quickcheck-instances`.

cc @RyanGlScott

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -1,5 +1,5 @@
 Name: QuickCheck
-Version: 2.9.2
+Version: 2.10
 Cabal-Version: >= 1.8
 Build-type: Simple
 License: BSD3
@@ -61,15 +61,8 @@ flag templateHaskell
   Description: Build Test.QuickCheck.All, which uses Template Haskell.
   Default: True
 
-flag light
-  Description:
-    Build with as few dependencies as possible.
-    Some instances will be missing, but build time will be reduced.
-    Can be useful on e.g. Travis.
-  Default: False
-
 library
-  Build-depends: base >=4.3 && <5, random, containers, array
+  Build-depends: base >=4.3 && <5, random, containers
 
   -- Modules that are always built.
   Exposed-Modules:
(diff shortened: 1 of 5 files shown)


phadej:
The build-times on travis seems to be quite quick!

QuickCheck.cabal
```diff
@@ -61,15 +61,8 @@ flag templateHaskell
...
-flag light
-  Description:
-    Build with as few dependencies as possible.
-    Some instances will be missing, but build time will be reduced.
-    Can be useful on e.g. Travis.
-  Default: False
-
 library
-  Build-depends: base >=4.3 && <5, random, containers, array
+  Build-depends: base >=4.3 && <5, random, containers
```
nick8325: Just a note: `containers` depends on `array` so we don't save anything by removing the array instances. On the other hand, since those instances were already in `quickcheck-instances` perhaps it's the right thing to do anyway.

Test/QuickCheck/Arbitrary.hs
```diff
@@ -166,26 +153,12 @@ import qualified Data.IntMap as IntMap
...
-#if MIN_VERSION_transformers(0,4,0)
-#define TRANSFORMERS_SUM
-#endif
-#else
-#define TRANSFORMERS_SUM
-#endif
-
-#ifdef TRANSFORMERS_SUM
-import Data.Functor.Sum
-#endif
```
nick8325: Shouldn't this be kept? It doesn't add any extra dependencies; `Data.Functor.Sum` simply doesn't exist if `transformers` is too old.

phadej: but it is in `transformers-compat`, and actually migrated to `base` recently.

So I went with simple and straight-forward no-CPP approach, as with `Natural` and `NonEmpty`.

nick8325: OK!

PR #167, opened 2017-05-24
Print classify percentages up to two significant digits

michalpalka:
Currently, the percentage values of occurrences of labels
in test cases are printed using integral percentages, which
is not precise enough if the labels occur very rarely. This
commit prints them up to two siginificant digits.

### DIFF
--- a/Test/QuickCheck/Test.hs
+++ b/Test/QuickCheck/Test.hs
@@ -32,6 +32,8 @@ import Data.Char
   ( isSpace
   )
 
+import Text.Printf (printf)
+
 import Data.List
   ( sort
   , group
@@ -352,7 +354,7 @@ success st =
  where
   allLabels = reverse
             . sort
-            . map (\ss -> (showP ((length ss * 100) `div` numSuccessTests st) ++ head ss))
+            . map (\ss -> (showP (labelPrecise (length ss) (numSuccessTests st)) ++ head ss))
             . group
             . sort
             $ [ concat (intersperse ", " s')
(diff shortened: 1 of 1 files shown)


michalpalka:
Actually, let me refactor this-I'll submit a new version soon.

Test/QuickCheck/Test.hs
```diff
@@ -361,20 +363,41 @@ success st =
               , not (null s')
               ]
 
-  covers = [ ("only " ++ show (labelPercentage l st) ++ "% " ++ l ++ ", not " ++ show reqP ++ "%")
+  covers = [ ("only " ++ show (labelPrecise (labelCount l st) (numSuccessTests st)) ++ "% " ++ l ++ ", not " ++ show reqP ++ "%")
            | (l, reqP) <- Map.toList (S.labels st)
            , labelPercentage l st < reqP
            ]
 
-  showP p = (if p < 10 then " " else "") ++ show p ++ "% "
```
michalpalka: I am not sure what was the purpose of this, but I left it as close to the original functionality as I could.

nick8325:
It seems to no longer print the labels in decreasing order of frequency.

I think it should be possible to do this simpler - I'll have a go at it.

sol:
How will this relate to the `labels` field of the `Result` type?  Does it represent the same information?  Will that still be the case after this PR?

I'm asking because it's type is `[(String, Int)]`, so I would have assumed we need to change that too, if we want to increase the precision.

nick8325:
Yes, changing the result type to match seems like a good idea...

PR #186, opened 2017-10-03
Use ShowS instead of String for constructing output (fixes #185)

### LINKED ISSUE #185: Construction of `output` is slow due to appending at the end of a string
Steps to reproduce:
```haskell
import Test.QuickCheck
import System.Clock

main :: IO ()
main = measure $ quickCheckResult (verbose prop) >>= print . length . output

prop :: Property
prop = verbose $ \n -> (n == (23 :: Int)) ==> True

measure :: IO a -> IO a
measure action = do
  t0 <- getTime Monotonic
  a <- action
  t1 <- getTime Monotonic
  putStrLn (show (toNanoSecs (t1 - t0) `div` 1000000) ++ "ms")
  return a
```

On my box this takes more than 5 seconds to complete.  If I don't look at the `output`, it's fast (< 500ms).  The faulty code is in `Test/QuickCheck/Text.hs`, `modifyIORef res (++ s)`.

sol:


sol:
@nick8325 can we get this on Hackage soon, please?

Test/QuickCheck/Text.hs
```diff
@@ -112,7 +112,7 @@ withNullTerminal action =
   newTerminal (const (return ())) (const (return ())) >>= action
 
 terminalOutput :: Terminal -> IO String
-terminalOutput (MkTerminal res _ _ _) = readIORef res
+terminalOutput (MkTerminal res _ _ _) = readIORef res >>= return . ($ "")
```
phadej: why not `fmap ($ "") (readIORef res)` ?

sol: Either way works for me, the only thing I care about is to avoid `<$>` for backwards compatibility.

nick8325:
Thanks!

nick8325:
2.10.1 is on Hackage now.

PR #191, opened 2017-12-23
Combinators for products

### LINKED ISSUE #155: Arbitrary combinators for simple types
I propose a concise way of creating random generators for simple enough types. The objective is the same as #112, but this does not use generics.

This should have been a PR but I couldn't follow all the conditions associated with backwards compatibility. It also uses some closed type families and overlapping instances in a way some may find too inelegant. So I am just going to put the proof of concept here, but I am willing to work on this if you are interested in integrating it into QuickCheck.

https://gist.github.com/Lysxia/ee12ab5364e97057a0b485c3fcbdba74

The conclusion is that we can define a function `arbitraryProduct` which is compatible with all of these types

    arbitraryProduct
      :: (Arbitrary a, Arbitrary b) => (a -> b -> c) -> Gen c
      :: _ => (a -> b -> c -> d) -> Gen d
      :: _ => (a -> b -> c -> d -> e) -> Gen e
      -- etc

We can then give a single constructor to the function, and it takes care of filling its fields.

    test :: Gen (Int, Double, Bool, String, Maybe Integer)
    test = arbitraryProduct (,,,,)

 There is a `(%)` combinator derived from this to cleanly generate sums with `frequency`:

    test2 :: Gen (Either Int Bool)
    test2 = frequency
      [ 50 % Left
      , 50 % Right
      ]

Lysxia:
Resolves #155

### DIFF
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -106,6 +106,8 @@ module Test.QuickCheck
   , shrink2
 
     -- ** Helper functions for implementing arbitrary
+  , arbitraryProduct2
+  , arbitraryProduct3
   , arbitrarySizedIntegral
   , arbitrarySizedNatural
   , arbitrarySizedFractional
--- a/Test/QuickCheck/Arbitrary.hs
+++ b/Test/QuickCheck/Arbitrary.hs
@@ -36,6 +36,8 @@ module Test.QuickCheck.Arbitrary
   , shrink2
 
   -- ** Helper functions for implementing arbitrary
+  , arbitraryProduct2
+  , arbitraryProduct3
   , arbitrarySizedIntegral        -- :: Integral a => Gen a
   , arbitrarySizedNatural         -- :: Integral a => Gen a
   , arbitraryBoundedIntegral      -- :: (Bounded a, Integral a) => Gen a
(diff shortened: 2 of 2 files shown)


nick8325:
Hmm, I like the functionality but I'm not sure about the name now (since it doesn't generate a product). I'm going to try and think of a better name (bikeshedding). Suggestions welcome!

Lysxia:
Sure! How about `arbitraryConstructor` or `arbitraryVariant`?

nick8325:
It can be used to apply an arbitrary function though, not just a constructor.

How about `liftArbitrary2` etc. by analogy with `liftM2` and `liftA2`?

phadej:
liftArbitrary is used by Arbitrary1, unfortunately

nick8325:
Ah, that's unfortunate. My first thought had been `arbitrary2`, which fails for the same reason :)

How about `arbitraryApply2`, then?

Lysxia:
I like `arbitraryApply2`!

nick8325:
I renamed to arbitraryApply2/3 and added an arity-4 version.

nick8325:
But I may do further bikeshedding if I think of a better name :)

Lysxia:
Great!

PR #194, opened 2017-12-26
Exported data constructor `Fun(Fun)`

asr:
Data constructor `Fun(Fun)` is not exported by `QuickCheck.hs`. This PR fix this issue.

### DIFF
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -158,7 +158,7 @@ module Test.QuickCheck
   , PrintableString(..)
 
     -- ** Functions
-  , Fun
+  , Fun (..)
   , applyFun
   , applyFun2
   , applyFun3


phadej:
Asr, have you noticed `Fn` pattern, it should be exported. IIRC we had good reason to not export Fun constructor, could you check a PR that added it to main module export list (should be recent, in the last year). But if there weren't any reason, than this is :+1:

asr:
@phadej, I hadn't notice the `Fn` pattern, thanks! 

From the example in the documentation 

```haskell
prop_foo :: Fun String Integer -> Bool
prop_foo (Fn f) = f "banana" == f "monkey" || f "banana" == f "elephant"
```

using the `Fn` pattern, I got the warning (when compiling with the `-Wall` flag):

```l
Pattern match(es) are non-exhaustive
    In an equation for ‘prop_foo’: Patterns not matched: _
```

Note that the warning disappears if I use the data constructor `Fun` instead:

```haskell
import Test.QuickCheck.Function

prop_bar :: Fun String Integer -> Bool
prop_bar (Fun _ f) = f "banana" == f "monkey" || f "banana" == f "elephant"
```

Can I avoid the above warning without adding the equation

```
prop_foo _ = ...
```

?

nick8325:
> IIRC we had good reason to not export Fun constructor, could you check a PR that added it to main module export list (should be recent, in the last year).

It's #104, but it just says "I intentionally didn't re-export `Fun` constructor". Anyway, seems like a good idea to add.

PR #195, opened 2018-01-05
Make (<*>) split only once

Lysxia:
`ap` calls `(>>=)` twice, and the third seed is discarded by `return`:

    ap mf mx = do f <- mf ; x <- mx ; return (f x)

Does this count as a breaking change? Does QuickCheck make stability guarantees that allow storing test cases as seeds?


nick8325:
There's no such stability guarantee - for example, generators often change between QuickCheck versions. So I think this patch is fine.

PR #197, opened 2018-01-12
Add Arbitrary instance for Natural

sestrella:


nick8325:
See #158 and #166. We don't provide this instance because on older GHC versions we would have to pull in the `nats` package, which would create a lot of extra dependencies. Instead, this instance is in the `quickcheck-instances` package.

sestrella:
@nick8325 thanks for the heads up!

PR #204, opened 2018-02-03
Add instance (Functor Fun)

Lysxia:


phadej:
Oh, now I'm tempted to add `Profunctor Fun` to `quickcheck-instances` (but `profunctors` aren't in Haskell Platform :/)

Lysxia:
I like the idea though!

PR #208, opened 2018-02-10
Function instances for Data.Monoid newtypes

Lysxia:


phadej:
`newtype Alt f a` *Since: 4.8.0.0*

Lysxia:
Okay it passes CI now. I hid the `Alt` instance for older `base` versions and also wrote explicit instances because older versions of `base` don't have `Generics`.

phadej:
LGTM

PR #216, opened 2018-06-14
Fix comment in MkGen

ppetr:
From looking at how the the parameter is used, it is obviously the size, not a seed.


nick8325:
Yes, there are two parameters, the seed and the size.

MaximilianAlgehed:
I've fixed this on `master`

PR #217, opened 2018-06-20
Add (=/=) inequality check 

tom-bop:


phadej:
`inspection-testing` uses `=/=` which is IMO nicer.

:+1: for the PR, I thought this one existed!

tom-bop:
After a minute of staring at the two, I agree!

nick8325:
Thanks!

PR #218, opened 2018-08-23
Add MonadFix instance for Gen

JonFowler:
Simple change adding a `MonadFix` instance for `Gen`

### DIFF
--- a/Test/QuickCheck/Gen.hs
+++ b/Test/QuickCheck/Gen.hs
@@ -21,6 +21,9 @@ import Control.Monad
   , filterM
   )
 
+import Control.Monad.Fix
+  ( MonadFix(..) )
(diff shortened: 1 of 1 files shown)


phadej:
I'm not a maintainer, but would be nice to see an example where this is useful (which could act as a test!)

JonFowler:
I found a better way of doing what I wanted to achieve so I'm closing this.

There is an argument that this should be in as it is a standard `base` typeclass and this is a valid instance but I don't know any good examples which aren't contrived.

nick8325:
I think this is a reasonable thing to have if you want to generate infinite structures.

I seems to satisfy the required laws, so I can merge this and add a little test case myself :)

PR #222, opened 2018-09-20
Introduced assertException and assertDeepException

lehins:
I found myself needing these functions on multiple occasions and thought that others might benefit from them as well. Please, let me know if there is anything you'd like me to change in order to get this PR merged.

Use case for those functions is pretty straightforward:
```haskell
λ> quickCheck $ \ x -> assertException (==DivideByZero) (x `div` 0)
+++ OK, passed 100 tests.
```

I thought at first `Test.QuickCheck.Exception` would be a good home for these functions, but it'll be too hard to avoid import cycles.

### DIFF
--- a/src/Test/QuickCheck/Monadic.hs
+++ b/src/Test/QuickCheck/Monadic.hs
@@ -72,6 +72,17 @@ module Test.QuickCheck.Monadic (
   , monadicST
   , runSTGen
 #endif
+
+  -- * Exceptions
+
+#ifndef NO_EXCEPTIONS
+  , assertException
+  , assertExceptionIO
+#ifndef NO_DEEPSEQ
+  , assertDeepException
+  , assertDeepExceptionIO
+#endif
+#endif
   ) where
 
 --------------------------------------------------------------------------
@@ -95,6 +106,13 @@ import Control.Monad.Trans.Class
 import qualified Control.Monad.Fail as Fail
 #endif
 
+#ifndef NO_DEEPSEQ
+import Control.DeepSeq
+#endif
+
+#ifndef NO_EXCEPTIONS
+import qualified Control.Exception as E
+#endif
 --------------------------------------------------------------------------
 -- type PropertyM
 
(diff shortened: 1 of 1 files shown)


MaximilianAlgehed:
Closes #265 and #230 as it would provide the necessary functionality for both of these.

Test/QuickCheck/Monadic.hs
```diff
@@ -273,5 +291,65 @@ runSTGen f = do
...
+
+#ifndef NO_DEEPSEQ
+
+-- | Evaluate the value to NF and check if it raises an exception.
+assertDeepException ::
+     (NFData a, E.Exception exc)
+  => (exc -> Bool) -- ^ Return True if that is the exception that was expected
+  -> a -- ^ Value that should throw an exception, when fully evaluated
+  -> Property
+assertDeepException isExc action = assertDeepExceptionIO isExc (return action)
```
MaximilianAlgehed, line 333: why not `assertDeepException isExc action = assertException isExc (rnf action)`?

Test/QuickCheck/Monadic.hs
```diff
@@ -273,5 +291,65 @@ runSTGen f = do
...
+
+-- | Make sure that an exception is thrown during an IO action. The result is
+-- evaluated to WHNF.
+assertDeepExceptionIO ::
+     (NFData a, E.Exception exc)
+  => (exc -> Bool) -- ^ Return True if that is the exception that was expected
+  -> IO a -- ^ An action that should throw the exception
+  -> Property
+assertDeepExceptionIO isExc action =
```
MaximilianAlgehed, line 342: Same as above?

MaximilianAlgehed (approved):

MaximilianAlgehed:
@lehins do you have the time to bring this up to date so we can merge it? If not just let me know and I can do it.

Thank you for the addition!

Test/QuickCheck/Monadic.hs
lehins: There is definitely no point in this code duplication. 
It's been some time since I wrote those functions, so I am not sure why I went with copy/paste solution :smile:

Test/QuickCheck/Monadic.hs
lehins: Done

Test/QuickCheck/Monadic.hs
MaximilianAlgehed: I did think it was a bit out of character! But it's a long time ago ;)

PR #223, opened 2018-09-29
Coarbitrary exists, even without generics

phadej:
@nick8325 it would be very great to have a `2.11.5` release with this.


nick8325:
Whoops. Thanks!

nick8325:
I've released 2.11.5 with your patch applied.

PR #235, opened 2018-11-27
Add HasCallStack where error can occur

dspies-leapyear:
When this fails due to an empty list there's no stack trace pointing to the bad "elements" call.

### DIFF
--- a/Test/QuickCheck/Gen.hs
+++ b/Test/QuickCheck/Gen.hs
@@ -35,6 +35,7 @@ import Test.QuickCheck.Random
 import Data.List
 import Data.Ord
 import Data.Maybe
+import GHC.Stack (HasCallStack)
 
 --------------------------------------------------------------------------
 -- ** Generator type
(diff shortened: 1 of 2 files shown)


Test/QuickCheck/Gen.hs
```diff
@@ -197,7 +198,7 @@ frequency xs0 = choose (1, tot) >>= (`pick` xs0)
   pick _ _  = error "QuickCheck.pick used with empty list"
 
 -- | Generates one of the given values. The input list must be non-empty.
-elements :: [a] -> Gen a
+elements :: HasCallStack => [a] -> Gen a
```
sol: This only works with recent versions of GHC.

My `call-stack` package provides a shim for backwards compatibility.  @nick8325 are you fine with adding that dependency?  If yes, we could also add source locations to e.g.  `===`.

If not, we would want to CPP the `HasCallStack => ` of the type signature.

dspies-leapyear: 👍

dspies-leapyear:
@sol The error message is certainly clear, but it's not useful for debugging if `elements` is called in a lot of places. For debugging purposes, I'd like to know _which_ call to `elements` is causing the failure without having to put `trace` calls everywhere or create my own `elements` wrapper (which is what I'm doing currently).

PR #237, opened 2018-12-12
Test.QuickCheck.Monadic: use 'pure' instead of 'assert' in docs

tom-audm:


tom-audm:
The rationale for this is that ```assert``` limits us to ```Bool``` types, but using ```pure``` is strictly more powerful (and lets us use ```Property```s, which are more idiomatic anyway).

MaximilianAlgehed:
This has gone stale by now and IMHO using `pure` to return something of the right type at the end of a monadic computation isn't something we need to document here.

Anyone who feels strongly about this is free to submit a new PR with updated docs that highlight how `PropertyM` works if they deem the current docs unclear.

PR #240, opened 2018-12-18
replace "Falsifiable" in error message into some better word

luochen1990:
I think use "Falsifiable" here is not very appropriate, since it emphasize the **inherent possibility** that a statement can be proven false. In our case, the proposition is already shown to be False, and we have found an obvious evidence about that, so I think "Falsism" could be better here.

> Falsifiability or refutability of a statement, hypothesis, or theory is the inherent possibility that it can be proven false. A statement is called falsifiable if it is possible to conceive of an observation or an argument which negates the statement in question.

> A falsism is a claim that is clearly and self-evidently wrong. A falsism is usually used merely as a reminder or as a rhetorical or literary device. An example is "pigs can fly." It is the opposite of truism.

### DIFF
--- a/Test/QuickCheck.hs
+++ b/Test/QuickCheck.hs
@@ -146,7 +146,7 @@ module Test.QuickCheck
     -- >>>     prop (Fun _ f) = f "monkey" == f "banana" || f "banana" == f "elephant"
     -- >>> :}
     -- >>> quickCheck prop
-    -- *** Failed! Falsism (after 3 tests and 134 shrinks):
+    -- *** Failed! Falsified (after 3 tests and 134 shrinks):
     -- {"elephant"->1, "monkey"->1, _->0}
     --
     -- To generate random values of type @'Fun' a b@,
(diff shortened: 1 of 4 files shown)


quasicomputational:
How about just 'Falsified'? 'Falsism' is a word I've literally never come across before today and I think most native speakers would be puzzled by it; it's got less than ten thousand Google hits, too.

luochen1990:
@quasicomputational Thanks for your suggestion, I think the semantic here is "this proposition have been proved to be false obviously", and I'm not sure which word express this semantic naturally for native speakers. Is there any other comment or suggestion?   should I change it to "Falsified" or something else?

Lysxia:
If I read the definition correctly, it's not a falsism if you need to prove it false, as is the case of QuickCheck properties, that look for counterexamples. Even my spell-checker doesn't know that word. "Falsified" sounds better.

luochen1990:
@Lysxia Yes, if you look at the total property, it is not an falsism, but this word appears in the following code:
```
liftBool :: Bool -> Result
liftBool True = succeeded
liftBool False = failed { reason = "Falsifiable" }
liftBool False = failed { reason = "Falsism" }
```
here, "falsism" explains why we return `failed` when we get a `False`.

But if this word "falsism" is so rarely used, I think it will not be a good idea to use it here.  I will change it to "Falsified".

nick8325:
"Falsified" sounds good to me. Thanks for the patch!

tom-audm:
For what it's worth, I prefer "Falsifiable", as in "your proposition is falsifiable, and here's an example that proves it"

"Failed! Falsified" feels repetitive to me.

PR #242, opened 2019-01-22
Update Gen.hs

amigalemming:
reference to new quickcheck-transformer package

### DIFF
--- a/Test/QuickCheck/Gen.hs
+++ b/Test/QuickCheck/Gen.hs
@@ -41,9 +41,11 @@ import Data.Maybe
 
 -- | A generator for values of type @a@.
 --
--- The third-party package
+-- The third-party packages
 -- <http://hackage.haskell.org/package/QuickCheck-GenT QuickCheck-GenT>
--- provides a monad transformer version of @GenT@.
+-- and
+-- <http://hackage.haskell.org/package/quickcheck-transformer quickcheck-transformer>
+-- provide monad transformer versions of @Gen@.
 newtype Gen a = MkGen{
   unGen :: QCGen -> Int -> a -- ^ Run the generator on a particular seed.
                              -- If you just want to get a random value out, consider using 'generate'.


phadej:
I wonder what's the difference between `quickcheck-GenT` and `quickcheck-transformer`? (And why there is virtually the same package twice on Hackage?)

amigalemming:
On Tue, 22 Jan 2019, Oleg Grenrus wrote:

> I wonder what's the difference between quickcheck-GenT and 
> quickcheck-transformer? (And why there is virtually the same package 
> twice on Hackage?)

As the description says: quickcheck-transformer is a fork of 
QuickCheck-GenT that does not need GHC extensions and works with GHC back 
to 7.0.4. The author of QuickCheck-GenT did not want maintain code for old 
GHC versions.

phadej:
Is there a point to mention `QuickCheck-GenT` at all then?

amigalemming:
On Tue, 22 Jan 2019, Oleg Grenrus wrote:

> Is there a point to mention QuickCheck-GenT at all then?

Not necessarily.

phadej:
FWIW, http://packdeps.haskellers.com/reverse/QuickCheck-GenT has only one reverse dependency. `quickcheck-transformer` seems to have none (yet?)

amigalemming:
On Tue, 22 Jan 2019, Oleg Grenrus wrote:

> FWIW, http://packdeps.haskellers.com/reverse/QuickCheck-GenT has only one reverse dependency.
> quickcheck-transformer seems to have none (yet?)

quickcheck-transformer is newer and thus has fewer users - but there is 
one:
   https://hackage.haskell.org/package/lapack-0.2.1/lapack.cabal

(only the test-suite depends on it)

nick8325:
I don't know if it's best to refer to both packages or only to `quickcheck-transformers`, but that's no reason to not merge the patch :)

PR #245, opened 2019-02-17
Negative & NonPositive

BebeSparkelSparkel:
Trying to keep from obfuscating my code by negating Positive.


phadej:
We'd need `NonPositive` for completeness too.

BebeSparkelSparkel:
@phadej I'm having problems building this. When I `cabal configure` it errors with `cabal: failed to parse output of 'ghc-pkg dump'`. What command should I use to build this?

Lysxia:
`cabal new-build` should be the only thing to build.

BebeSparkelSparkel:
@Lysxia Thanks. I needed to reinstall cabal

nick8325:
Thanks! In the end I rewrote the patch to generate negative numbers directly, because negating the output of `Positive` wouldn't ever generate `minBound` for twos-complement integers.

BebeSparkelSparkel:
I didn't even think of that. Appreciate your help getting that right

nick8325:
I think that `Positive` had a similar bug in the past, which is why I spotted it :)

PR #247, opened 2019-03-12
cchoose - curried choose

BebeSparkelSparkel:


BebeSparkelSparkel:
I am finding that I am using `(curry choose)` far more than `choose` and it would be nice type less.

nick8325:
I think that `curry choose` is easier to understand than `cchoose` (especially if you don't see it often), so I'm not convinced that it's worth adding an extra combinator to the API for this.

I do think it would have been nicer if `choose` had been curried from the beginning though.

PR #251, opened 2019-03-27
added or

BebeSparkelSparkel:


phadej:
`or` is a name in `Prelude`. :-1:

MaximilianAlgehed:
I think this is two spaces shorter than writing `elements [x, y]`. Let's not bloat the API for little to no gain.

PR #255, opened 2019-04-27
Fix GHC 8.8 build: remove fail from Monad instance for newer GHCs

Bodigrim:


phadej:
Conditional should be on MIN_VERSION_base(4,13,0)

Bodigrim:
> Conditional should be on MIN_VERSION_base(4,13,0)

To be 100% compatible - yes. But it makes the logic of conditionals harder to grasp and is out of line with the rest of the module. AFAIU as long as all usages of `PropertyM` goes through `monadicIO` / `assert` pattern, it does not change semantics.

phadej:
It does change the public interface, and as you propose it, it’s breaking change.

`fail` records the reason (`assert` only says ”Assertion failed”), so I’d expect it’s used to greater extend in the users testing code.

Bodigrim:
Agreed, makes sense.

nick8325:
Thanks!

PR #260, opened 2019-05-22
Add bounds for random

phadej:
- Lower bound: prior random-1.0.1.0 it wasn't `Trustworthy`, so compiling QuickCheck fails
- Upper bound: `random-1.2` (or whatever next major) will change things, though it's not clear when it will be released. Better to have bounds in place, so we won't need to make revisions later.


nick8325:
Thanks!

Issue #257, opened 2019-05-09
Way to write `Function Void` instance.


phadej:
It would be trivial instance:

```haskell
-- | All @'Void' -> a@ are 'absurd'.
instance Function Void where
    function _ = Nil
```

but `:->` constructors aren't exposed from anywhere.

Should they be, from some internal module?

nick8325:
This instance seems like a good idea.

I don't think there's a need to export the definition `:->` as `functionMap` can always be used to define new instances, as long as there are instances for sums, products, `()` and `Void`.

phadej:
But `Void` isn't in old `base`s. That's the problem. Is adding

```haskell
functionVoid :: a :-> c
functionVoid = Nil
```

be ok?

Then I can add `Function Void` for all supported GHCs and `base` into `quickcheck-instances`.

nick8325:
Sorry for the slow reply.

I'd rather not commit to making `Nil` have the type `a :-> b` for all time. Could we make `functionVoid` take an extra argument that witnesses that the type is void, something like:

```
functionVoid :: (forall b. a -> b) -> a :-> c
functionVoid _ = Nil
```

You would then call it as `functionVoid absurd`. It won't make it impossible for people to use `functionVoid` at non-Void types, but it might make them think twice.

PR #269, opened 2019-11-01
`CoArbitrary` instance for Data.Void.

leftaroundabout:
The vacuous type `Void` contains no values. It can thus not have an `Arbitrary` instance, but it does offer a (trivial) `CoArbitrary`.
While this is not really useful per se, it can be used to "disable" some parametric field of an alternative type.

### DIFF
--- a/Test/QuickCheck/Arbitrary.hs
+++ b/Test/QuickCheck/Arbitrary.hs
@@ -86,6 +86,7 @@ module Test.QuickCheck.Arbitrary
 
 import Control.Applicative
 import Data.Foldable(toList)
+import Data.Void (Void, absurd)
 import System.Random(Random)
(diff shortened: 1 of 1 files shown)


phadej:
https://hackage.haskell.org/package/quickcheck-instances-0.3.22/docs/Test-QuickCheck-Instances-Void.html

`Void` is not in `QuickCheck` as its not in all `base` supported by `QuickCheck`

leftaroundabout:
Ah. I thought such a simple instance would surely be defined right there in the default package and was just forgotten because it's normally useless, but if it's in the instance package, fine.

phadej:
For the reference see also https://github.com/nick8325/quickcheck/issues/257

PR #274, opened 2019-11-08
Fix non-valid unicode issue

boris-stepanov:
By standard, Unicode has sixty-six noncharacters, which can be incorrectly
interpreted by different applications. To prevent possible troubles,
characters with NotAssigned property shouldn't be generated.


nick8325:
Thanks!

PR #281, opened 2019-11-20
Improve Gen instances

treeowl:
* Write custom `(*>)`, `(<*)`, and `(>>)` implementations for `Gen`;
  these can avoid splitting the seed. They're also essentially useless,
  but they could show up if someone uses `Applicative`-generic
  operations of some sort.

* Move the implementation from `return` to `pure` for `Gen`, rendering
  the `return` implementation "canonical".


nick8325:
Thanks!

PR #283, opened 2019-12-02
Export functionEitherWith and functionPairWith

phadej:
These allows constructing `:->` without looping though `Function` type-class


phadej:
ping

PR #290, opened 2020-03-22
Export RecursivelyShrink & GSubterms

tmortiboy:
I am using `genericShrink` but require `RecursivelyShrink` and `GSubterms` to be exported for my type signatures.

Would you be open to exporting these types?


nick8325:
Thanks!

PR #291, opened 2020-04-09
added applicative to num wrappers

BebeSparkelSparkel:


Lysxia:
These instances seem weird to me, including the already existing `Functor`s. We shouldn't add them "just because we can". Do you have a use case for these?

My objection is that it's fishy that we can even write `fmap negate (Positive 1)`, which equals `Positive (-1)`. The constructor might be exposed anyway, but that is trumped by the convenience of the way they're actually used, by pattern-matching in a lambda: `\(Positive x) -> ...`.

Is there another use for these wrappers that makes sense?

BebeSparkelSparkel:
It seemed convenient to be able to do
```haskell
(|-|) :: (Applicative f, Num a) => f a -> f a -> f a
(|-|) = liftA2 (-)
```

MaximilianAlgehed:
There appears to be no clear use-case for this and the PR has been stale for several years. I'm going to close this but feel free to open this PR again with a compelling example of where to use this and we can have another look!

PR #297, opened 2020-05-16
Tweak Float and Double Arbitrary Instances.

phadej:
Tries to address comments in https://github.com/nick8325/quickcheck/issues/295
and https://github.com/nick8325/quickcheck/pull/296

Doesn't change how arbitrarySizedFractional works.
Now it is only used for (Ratio a) instance.

Also two properties which should fail everytime:
the Double generator has a chance of generating same values,
and it also hits values which break addition associativity.

---

My take on #295

### DIFF
--- a/src/Test/QuickCheck/Arbitrary.hs
+++ b/src/Test/QuickCheck/Arbitrary.hs
@@ -683,11 +683,55 @@ instance Arbitrary Char where
                 )
 
 instance Arbitrary Float where
-  arbitrary = arbitrarySizedFractional
+  arbitrary = oneof
+    -- generate 0..1 numbers with full precision
+    [ genFloat
+    -- generate integral numbers
+    , fromIntegral <$> (arbitrary :: Gen Int)
+    -- generate fractions with small denominators
+    , smallDenominators
+    -- uniform -size..size with with denominators ~ size
+    , uniform
+    -- and uniform -size..size with higher precision
+    , arbitrarySizedFractional
+    ]
+    where
+      smallDenominators = sized $ \n -> do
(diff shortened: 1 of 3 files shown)


cartazio:
Looks like at least a nice starting point.  Cc @Bodigrim

cartazio:
Thx!

cartazio:
Does this code generate negative zero?

phadej:
> Does this code generate negative zero?

It doesn't. I'll add it.

cartazio:
I’d also advocate for plus and minus infinity. Since those sometimes have semantic meaning. Ish. And or the smallest and largest magnitude finite floats. 

Nans don’t matter / should be runtime exceptions ;)

phadej:
With infinities you easily get yourself into NaN-land, at which point almost every property is false. That feels like arbitrary choice, who's life you want to make easier/harder
- People testing numeric code (who probably know what they are doing)
- People with some `Double`s in their business data types (who probably know less well, to not use `Double` there).

```haskell
(1/0.0 :: Double) - (1/0.0 :: Double)
NaN
```

cartazio:
Touche.  Either way nontrivial chance of zeros or duplicates will go a long
way!

For a lotta applied math codes, random all distinct numbers in floating
point guarantees the code will work almost too well.  :)

cartazio:
i emphatically support this change, it seems like it will help MANY people

phadej:
Resolved conflict.

phadej:
@nick8325 ping

nick8325:
Sorry for my extremely slow response to this patch - it looks really useful! Merged now.

PR #303, opened 2020-06-23
Compatibility with new version of random

lehins:
This PR implements compatibility with `random-1.2.0`, which was finally released today

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -1,5 +1,5 @@
 Name: QuickCheck
-Version: 2.14
+Version: 2.14.1
 Cabal-Version: >= 1.8
 Build-type: Simple
 License: BSD3
@@ -63,7 +63,7 @@ flag templateHaskell
   Default: True
 
 library
-  Build-depends: base >=4.3 && <5, random >=1.0.0.3 && <1.2, containers
+  Build-depends: base >=4.3 && <5, random >=1.0.0.3 && <1.3, containers
 
   -- random is explicitly Trustworthy since 1.0.1.0
   -- similar constraint for containers
--- a/Test/QuickCheck/Random.hs
+++ b/Test/QuickCheck/Random.hs
@@ -11,6 +11,9 @@ import System.Random
 import System.Random.SplitMix
 #endif
 import Data.Bits
+#if MIN_VERSION_random(1,2,0)
+import Data.Coerce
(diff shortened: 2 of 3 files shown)


NorfairKing:
This promises amazing performance gains. I'm looking forward to it!

phadej:
@NorfairKing `QuickCheck` is already using `splitmix` and not using `random` for about anything, so I wouldn't expect much more speedups than what 2.14 already brought to the table. See also #299 (which I'd prefer this PR to based on).

Also I don't see anything truly breaking here, thus I'm not sure `2.15` bump is justified.

lehins:
@phadej switched to 0.14.1

>  wouldn't expect much more speedups than what 2.14 already brought to the table

Looks like all you had to do was wait for the new version of random to be released, instead now there is `NO_SPLITMIX` all over the place.

Any reconciliation `choose` function will get faster ;)

phadej:
`NO_SPLITMIX` is new `NO_TF_RANDOM`, which [was introduced in 2014](https://github.com/nick8325/quickcheck/commit/59bb2eeb) ;) And I think the research behind `tf-random` was motivated by `QuickCheck` use case.

lehins:
Yeah, I remember about `tf-random`, I am not sure what happened, but somehow it got stalled. Thank you for implementing the `splitmix` package, otherwise we'd still be stuck with outdated RNGs.

In any case, now that `random-1.2` is out `NO_SPLITMIX` seems silly:

```haskell
#ifdef NO_SPLITMIX
newtype QCGen = QCGen StdGen
#else
newtype QCGen = QCGen SMGen
#endif
```

Considering the new definition of `StdGen`:

```haskell
newtype StdGen = StdGen { unStdGen :: SMGen }
```

It's up to you, of course, what you are going to do about it, if anything at all. Matters to me not.

phadej:
```haskell
#ifdef NO_SPLITMIX
newtype QCGen = QCGen StdGen
#else
newtype QCGen = QCGen SMGen
#endif
```

matters for GHCs older than 8.2. And because `QuickCheck` is testing library, I think it's justified to do a little of CPP to extend support window by a lot.

lehins:
@phadej FYI, we extended support to ghc>=7.10 in random-1.2

phadej:
@lehins good to know, thanks!

PR #306, opened 2020-06-26
Update quickcheck to build with GHC head (8.11).

facundominguez:


nick8325:
I ended up merging #314 instead, which is very similar. Thanks for the patch!

PR #308, opened 2020-07-23
Add Tree instances

phadej:
Code taken from https://github.com/phadej/qc-instances/pull/30
(sorry it took so long to come to it).

Small change is to use shuffle in arbPartition so they
won't be "tail heavy".

Also add Function instance which was missing from
quickcheck-instances


phadej:
GHC-7.8.4 failure seems to be transient.

phadej:
ping

nick8325:
Looks good. Thanks!

PR #311, opened 2020-07-28
Support GHC-9.0 (template-haskell-2.17)

phadej:
Patch is originally by @RyanGlScott
https://gitlab.haskell.org/ghc/head.hackage/-/blob/0ffef791971bde9a519762719cbdad570ac446f1/patches/QuickCheck-2.14.1.patch


buggymcbugfix:
I believe this patch is not good. See #314.

PR #314, opened 2020-08-24
Support GHC-9.0 (template-haskell-2.17) and allow `StarT`-kinded TVs

buggymcbugfix:
(1) Fix bad unipattern in lambda
(2) Update patterns for updated `TyVarBndr` from template-haskell-2.17.0

While opening this PR I noticed that #311 also addresses (2), however it addresses (1) differently and apparently incorrectly.

The uni-patterned lambda would previously incorrectly fail on `KindedTV nm StarT` and #311 unnecessarily reproduces this failure.

This patch is also shorter and clearer.

### DIFF
--- a/src/Test/QuickCheck/All.hs
+++ b/src/Test/QuickCheck/All.hs
@@ -99,11 +99,16 @@ infoType (VarI _ ty _ _) = ty
 
 deconstructType :: Error -> Type -> Q ([Name], Cxt, Type)
 deconstructType err (ForallT xs ctx ty) = do
-  let plain (PlainTV nm)          = return nm
-#if MIN_VERSION_template_haskell(2,8,0)
-      plain (KindedTV nm StarT)   = return nm
+#if MIN_VERSION_template_haskell(2,17,0)
+  let plain (PlainTV nm _)        = return nm
+      plain (KindedTV nm _ StarT) = return nm
 #else
+  let plain (PlainTV nm)          = return nm
+#  if MIN_VERSION_template_haskell(2,8,0)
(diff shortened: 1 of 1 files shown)


phadej:
ping @RyanGlScott

RyanGlScott:
Is there something in particular you want me to look at? Keep in mind that I'm not a maintainer of this library, so I don't really understand what this code even does at present.

phadej:
@RyanGlScott does the changes between your patch and this one make sense to you?

RyanGlScott:
I guess? Admittedly, I wrote that patch as quickly as possible to get the Hackage ecosystem compiling on HEAD, and I even preserved the semantics of the partial uni-pattern. (Again, it's worth emphasizing that I don't know what this code in `QuickCheck` is doing.)

Generally speaking, I like code without partial uni-patterns, so if this patch gets rid of that, then I'm all for it.

buggymcbugfix:
In that case, do you want to close #311 to avoid confusion, @phadej?

galenhuntington:
It looks like a `#endif` got accidentally inserted here, which is why the tests are failing:
https://github.com/nick8325/quickcheck/pull/314/commits/b45351fc824a5a2c16312b731749ac7983a4deac#diff-a2ca13afe67e562cd605374d2e9bb810L108-R114

buggymcbugfix:
@galenhuntington good spot, thanks. Should be fixed now.

nick8325:
Thanks a lot!

ethercrow:
@nick8325 Hi! Could you please release this on hackage?

nick8325:
Done!

Issue #322, opened 2020-11-24
Missing Arbitrary Instance for NewlineMode


dbramucci:
I noticed that there are no `Arbitrary` or `CoArbitrary` instances for the [`Newline`](https://hackage.haskell.org/package/base-4.14.0.0/docs/System-IO.html#t:Newline) and [`NewlineMode`](https://hackage.haskell.org/package/base-4.14.0.0/docs/System-IO.html#t:NewlineMode) types.

I encountered this missing instance while writing code for the `ByteString` package intended to test how `hGetLine` behaves on all platforms. The particular use-case is [this property](https://github.com/haskell/bytestring/blob/a3d4c5fe3ce62ec6da67b8c3c3ddab8f4a68a800/tests/Properties.hs#L1617) where I work around the issue by grabbing two `Bool`s instead of an arbitrary `NewlineMode`.

PR #323, opened 2020-11-27
Implemented instances for Newline and NewlineMode

dbramucci:
These are similar to the instances for bools and a pair of bools *but* we shrink towards `LF` because, according to `Newline`'s [documentation in base](https://hackage.haskell.org/package/base-4.14.0.0/docs/System-IO.html#g:25), Haskell assumes `\n` in its programs, and `LF` means we don't need any translations.
`CRLF`, means we do need to translate, meaning it will normally be the more complex case.

By implementing `Arbitrary`, `CoArbitrary` and `Function` instances for both [`Newline`](https://hackage.haskell.org/package/base-4.14.0.0/docs/System-IO.html#t:Newline) and [`NewlineMode`](https://hackage.haskell.org/package/base-4.14.0.0/docs/System-IO.html#t:NewlineMode), this closes issue #322

### DIFF
--- a/src/Test/QuickCheck/Arbitrary.hs
+++ b/src/Test/QuickCheck/Arbitrary.hs
@@ -138,7 +138,7 @@ import Data.List
 import Data.Version (Version (..))
 
 #if defined(MIN_VERSION_base)
-#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,4,0) && !defined(__NHC__) && !defined(__HUGS__)
+#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,2,0) && !defined(__NHC__) && !defined(__HUGS__)
 import System.IO
   ( Newline(..)
   , NewlineMode(..)
@@ -997,7 +997,7 @@ instance Arbitrary ExitCode where
   shrink _        = []
 
 #if defined(MIN_VERSION_base)
-#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,4,0) && !defined(__NHC__) && !defined(__HUGS__)
+#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,2,0) && !defined(__NHC__) && !defined(__HUGS__)
 instance Arbitrary Newline where
   arbitrary = elements [LF, CRLF]
 
@@ -1475,7 +1475,7 @@ instance CoArbitrary Version where
   coarbitrary (Version a b) = coarbitrary (a, b)
 
 #if defined(MIN_VERSION_base)
-#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,4,0) && !defined(__NHC__) && !defined(__HUGS__)
+#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,2,0) && !defined(__NHC__) && !defined(__HUGS__)
 instance CoArbitrary Newline where
   coarbitrary LF = variant 0
   coarbitrary CRLF = variant 1
--- a/src/Test/QuickCheck/Arbitrary.hs
+++ b/src/Test/QuickCheck/Arbitrary.hs
@@ -137,10 +137,14 @@ import Data.List
 
 import Data.Version (Version (..))
 
+#if defined(MIN_VERSION_base)
+#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,4,0) && !defined(__NHC__) && !defined(__HUGS__)
 import System.IO
   ( Newline(..)
   , NewlineMode(..)
   )
+#endif
+#endif
 
 import Control.Monad
   ( liftM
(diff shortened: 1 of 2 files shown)


src/Test/QuickCheck/Arbitrary.hs
```diff
@@ -1447,6 +1474,17 @@ instance CoArbitrary (f a) => CoArbitrary (Monoid.Alt f a) where
 instance CoArbitrary Version where
   coarbitrary (Version a b) = coarbitrary (a, b)
 
+#if defined(MIN_VERSION_base)
+#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,4,0) && !defined(__NHC__) && !defined(__HUGS__)
```
phadej, line 1478: Docs say these are available since base 4.3.0.0

https://hackage.haskell.org/package/base-4.3.0.0/docs/System-IO.html#t:Newline, i.e. on all `base` `QuickCheck` supports  - only not hugs.

dbramucci: I thought I checked 4.3 but apparently I didn't.
I just realized that the versions shuffle around as you go through them on Haddock.

Screenshot_2020-11-27 base Basic libraries(1)

Screenshot_2020-11-27 base Basic libraries(2)

So I probably thought I had checked 4.3 when I checked 4.2.0.2, jumped ahead in my search and saw that 4.3 was the last version on line 1.

Anyways, it is defined in `GHC/IO/Handles/Types.hs` and get's re-exported from `System/IO.hs` with the following code

```haskell
#if !defined(__NHC__) && !defined(__HUGS__)
    -- * Newline conversion
    
    -- | In Haskell, a newline is always represented by the character
    -- '\n'.  However, in files and external character streams, a
    -- newline may be represented by another character sequence, such
    -- as '\r\n'.
    --
    -- A text-mode 'Handle' has an associated 'NewlineMode' that
    -- specifies how to transate newline characters.  The
    -- 'NewlineMode' specifies the input and output translation
    -- separately, so that for instance you can translate '\r\n'
    -- to '\n' on input, but leave newlines as '\n' on output.
    --
    -- The default 'NewlineMode' for a 'Handle' is
    -- 'nativeNewlineMode', which does no translation on Unix systems,
    -- but translates '\r\n' to '\n' and back on Windows.
    --
    -- Binary-mode 'Handle's do no newline translation at all.
    --
    hSetNewlineMode, 
    Newline(..), nativeNewline, 
    NewlineMode(..), 
    noNewlineTranslation, universalNewlineMode, nativeNewlineMode,
#endif
```

Until version 4.7 where the `#if !defined(__NHC__) && !defined(__HUGS__)` was dropped.
And that's why I repeated the check for versions 4.4 <= x < 4.7.
But, that should be 4.3 not 4.4.

dbramucci: I just realized that I may have misinterpreted what you wrote, do you mean that because `QuickCheck` depends on `base >= 4.3`
I should simplify that line to something like

```haskell
#if MIN_VERSION_base(4,7,0) || !defined(__NHC__) && !defined(__HUGS__)
```

Or something completely different like

```haskell
#if defined(__GLASGOW_HASKELL__)
```

phadej: I'd use latter. It's way simpler.

dbramucci: I suppose

```haskell
#if defined(MIN_VERSION_base)
#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,2,0) && !defined(__NHC__) && !defined(__HUGS__)
```

Could be simplified to


```haskell
#if defined(MIN_VERSION_base)
```

Which would only cause trouble if one was using either a `base` without `QuickCheck` support *or* using `base` with `NHC` or `HUGS` instead of `GHC` for versions of `base` before 4.2.

The major hangup I have is that

```haskell
#if defined(MIN_VERSION_base)
#if MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,2,0) && !defined(__NHC__) && !defined(__HUGS__)
```

or

```haskell
#if defined(MIN_VERSION_base) && (MIN_VERSION_base(4,7,0) || MIN_VERSION_base(4,2,0) && !defined(__NHC__) && !defined(__HUGS__))
```

Tells me that the reason why this is `#if`ed in the first place is

* Something was added in base 4.2.0 that required `!defined(__NHC__) && !defined(__HUGS__)` and
* The `!defined(__NHC__) && !defined(__HUGS__)` requirement was dropped in base 4.7.0.

Whereas

```haskell
#if defined(MIN_VERSION_base)
```

```haskell
#if defined(__GLASGOW_HASKELL__)
```

Vaguely hint at something GHC specific but honestly, I'd need to `git blame` and find this pull request to understand why this was `#if`ed in the first place.

dbramucci: Furthermore, I honestly don't know the non-GHC world of Haskell well and the long macro is the only one that is "obviously correct" to my eyes.
It depends on certain versions of `base` and the status of the `__HUGS__` and `__NHC__` so I trust the long macro.

`#if defined(MIN_VERSION_base)` makes me worry about whether the hypothetical HUGS user on base 4.4 exists and `#if defined(__GLASGOW_HASKELL__)` isn't something I have a good logical argument for being correct.

Bodigrim: To be honest, non-GHC world of Haskell is barely existing. Even if you wish to be pedantic, `#if defined(__GLASGOW_HASKELL__)` is more than enough.

PR #325, opened 2021-01-08
Fix installation instructions in README

dpwiz:
The `cabal install` way suggested 10+ years ago before the advent of sandboxed projects.

### DIFF
--- a/README
+++ b/README
@@ -1,8 +1,6 @@
 This is QuickCheck 2, a library for random testing of program properties.
 
-Install it in the usual way:
-
-$ cabal install
+Add `QuickCheck` to your package dependencies to use it in tests or REPL.
 
 The quickcheck-instances [1] companion package provides instances for types in
 Haskell Platform packages at the cost of additional dependencies.


danidiaz:
There is a steady stream of questions in forums and Stack Overflow about failed attempts to install QuickCheck, due to the obsolete instructions: [here](https://stackoverflow.com/questions/63837574/haskell-could-not-find-module-test-quickcheck) [here](https://www.reddit.com/r/haskell/comments/ksz2iz/how_to_install_quickcheck_on_windows/) [here](https://www.reddit.com/r/haskell/comments/jdgfui/quickcheck_cant_be_found/) [here](https://www.reddit.com/r/haskell/comments/ivesrj/i_am_trying_to_install_quickcheck_using_cabal_but/) [here](https://www.reddit.com/r/haskell/comments/io56o3/could_not_find_module_testquickcheck_on_windows/) [here](https://stackoverflow.com/questions/52394677/how-to-install-quickcheck-on-windows).

The instructions are obsolete because Cabal has [moved on](https://cabal.readthedocs.io/en/3.4/nix-local-build-overview.html?highlight=nix-like) from destructive reinstalls (the "Cabal hell" of earlier times) and also from the use of sandboxes.

Now `cabal install` doesn't even install libraries by default, one needs to use [`cabal install --lib`](https://cabal.readthedocs.io/en/3.4/cabal-commands.html#cabal-v2-install) for that. That installs the library and registers it on a global [package environment file](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/packages.html#package-environments). One can also create package environments local to the current folder with `cabal install --lib --package-env .` (which I tend to prefer).

However `cabal install --lib` will likely be superseded by a [specialized `cabal env` subcommand](https://github.com/haskell/cabal/issues/6481)  in future versions of Cabal.

I would expand the installation instructions, giving two options:

- `cabal install --lib --package-env . QuickCheck` for quick and dirty experimentation using with ghci in the current folder.  (Note: also mention the need to `import Data.QuickCheck` in ghci).
- For robustness and reproducibility, explain how to create a [minimal cabal project](https://medium.com/@danidiaz/minimal-cabal-files-revisited-a5810215b0d9) with QuickCheck as a dependency, suggesting also the use [`cabal build`](https://cabal.readthedocs.io/en/3.4/cabal-commands.html#cabal-v2-build) / [`cabal repl`](https://cabal.readthedocs.io/en/3.4/cabal-commands.html#cabal-v2-repl) / [`cabal run`](https://cabal.readthedocs.io/en/3.4/cabal-commands.html#cabal-v2-run) commands.

PR #335, opened 2021-08-16
Prepare for monomorphic Data.List

Bodigrim:
Future GHCs will monomorphise `Data.List`, returning its API to pre-FTP times and encouraging qualified import of `Data.List`, similar to other containers. See [MR 5304](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5304) for more information. While this change has not been released yet, it already affects quality of life of folks testing GHC HEAD. The proposed patch is rather non-intrusive and backwards-compatible, so I suggest it would not hurt to merge it early.

CC @phadej

### DIFF
--- a/src/Test/QuickCheck/All.hs
+++ b/src/Test/QuickCheck/All.hs
@@ -29,7 +29,7 @@ import Language.Haskell.TH
 import Test.QuickCheck.Property hiding (Result)
 import Test.QuickCheck.Test
 import Data.Char
-import Data.List
+import Data.List (isPrefixOf, nubBy)
 import Control.Monad
 
 import qualified System.IO as S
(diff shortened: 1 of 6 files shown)


src/Test/QuickCheck/All.hs
```diff
@@ -29,7 +29,7 @@ import Language.Haskell.TH
 import Test.QuickCheck.Property hiding (Result)
 import Test.QuickCheck.Test
 import Data.Char
-import Data.List
+import Data.List as List
```
phadej, line 32: This is not correct  way to fix this compability, rather you should use

```
import Data.List (nubBy, isPrefixOf)
```

The same applies to other modules. Use explicit import list.

phadej (changes requested):
I thought that the change will be reverted and CLC will post a some "how to fix your code" message.

src/Test/QuickCheck/All.hs
Bodigrim: Thanks!

Bodigrim:
@phadej The change got delayed, but still in GHC HEAD and on the CLC road map in general. As you might remember from our discussion, I see no point in delaying.

nick8325:
Thanks!

PR #357, opened 2023-05-18
Speed up Test.QuickCheck.Text.putTemp

Bodigrim:
`Test.QuickCheck.Text.putTemp` is used a tight loop, updating progress after each test iteration. For a cheap property `putTemp` itself can contribute significantly to the execution time. The new implementation emits roughly twice less characters and thus makes this overhead slightly less pronounced.

@nick8325 it would be great if this can fit into the next release.


coot (approved):

nick8325:
Thanks!

PR #365, opened 2024-03-15
Introduce `withNumTests` and deprecate `withMaxSuccess`

MaximilianAlgehed:
After a discussion about changing the interaction between `checkCoverage` and `withMaxSuccess` whereby `checkCoverage` will ignore the number given in `withMaxSuccess` (because `checkCoverage` may require _more_ tests than what is specified) it was decided that a more intuitive name for `withMaxSuccess` is `withNumTests`. The new name reflects the fact that `withNumTests` does not set a hard limit (unlike what is indicated by the `Max` in `withMaxSuccess`) on the number of tests that are run.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,6 @@
+UNRELEASED
+	* Deprecate `withMaxSuccess` in favour of the renamed `withNumTests`
+
 QuickCheck 2.14.3 (released 2023-05-31)
 	* Add shrinkBoundedEnum (thanks to Jonathan Knowles)
 	* Add discardAfter for discarding tests on timeout (thanks to Justus Sagemüller)
--- a/src/Test/QuickCheck.hs
+++ b/src/Test/QuickCheck.hs
@@ -19,9 +19,9 @@ You can then use QuickCheck to test @prop_reverse@ on 100 random lists:
 >>> quickCheck prop_reverse
 +++ OK, passed 100 tests.
 
-To run more tests you can use the 'withMaxSuccess' combinator:
+To run more tests you can use the 'withNumTests' combinator:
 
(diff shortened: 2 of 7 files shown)


MaximilianAlgehed:
I tried to get a grip on the number of people we will annoy with this change by searching for the following query on github `withMaxSuccess language:haskell  NOT is:archived NOT is:fork` and it gave back ~600 files (with at most 50 files being instances of `withMaxSuccess` used in a comment).

Now, the thing I wasn't able to ascertain was how many distinct repositories would be impacted, but would I suspect it's somewhere between 60 and 300.

MaximilianAlgehed:
The reason for making this change is the change in #383 that means that `maxSuccess` is never really _max_ success (it wasn't before either - but now it's even more so not the case). Weighing that against pissing up to 600 people off however doesn't immediately worth while to me.

My plan is to wait with this change until we have enough other changes that will annoy people that we _can't_ ignore and just bundle all the breaking changes in one place.

nick8325:
Another option would be to introduce `withNumTests` but not yet mark `withMaxSuccess` as `{-# DEPRECATED #-}`. We could still update the Haddocks to tell people to use `withNumTests` instead.

phadej:
@MaximilianAlgehed https://hackage-search.serokell.io/ is better corpus to search through

MaximilianAlgehed:
> @MaximilianAlgehed https://hackage-search.serokell.io/ is better corpus to search through

Thank you! I'll do a more thorough search there.

MaximilianAlgehed:
There are about 50 odd packages on hackage that use `withMaxSuccess` that we have to care about (i.e. aren't mentioning it only in a comment).

PR #368, opened 2024-03-19
Introduce HasCallStack on new GHC versions

MaximilianAlgehed:
Closes #235


nick8325:
I suppose we could avoid duplicating the type signatures by doing something like:

```
#ifndef NO_CALLSTACK
import GHC.Stack
#define WITHCALLSTACK(ty) HasCallStack => ty
#else
#define WITHCALLSTACK(ty) ty
#endif
```

and then the type signatures would just be e.g. `resize :: WITHCALLSTACK(Int -> Gen a -> Gen a)`.

MaximilianAlgehed:
That's a good idea. I was thinking about doing something like `type WithCallStack a = a` or similar but I wasn't confident that no compiler would complain about that (though they shouldn't?!)

MaximilianAlgehed:
But the macro idea is nicer because it won't show up in the Haddocks.

UlfNorell (approved):

PR #371, opened 2024-03-20
Fix discard ratio and maxSuccess interacting poorly

### LINKED ISSUE #338: quickCheckWithResult with low maxSuccess fails unexpectedly
Contrary to expectation, lowering the `maxSuccess` parameter may cause tests to fail:

```
> let p k k' = k /= k' ==> (k :: Int) /= k'
> quickCheckWithResult stdArgs p
+++ OK, passed 100 tests; 11 discarded.
> quickCheckWithResult stdArgs{maxSuccess=2} p
+++ OK, passed 2 tests; 11 discarded.
> quickCheckWithResult stdArgs{maxSuccess=1} p
*** Gave up! Passed only 0 tests; 10 discarded tests.
```

The explanation seems to be due to a combination of:
- discard ratio defaults to 10, which means we allow only 10 discarded results in total for the case `maxSuccess=1` (`maxSuccess=2` gets 20 discards upfront)
- `computeSize` doesn't ramp up the size quickly enough, it starts out `[0,0,0,0,0,0,0,0,0,1,1,1,1,1,1,1,1,1,1,2,2,2...` regardless of `maxSuccess` parameter

Those computed sizes in particular seem a bit weird to me, but I can't say I have any kind of understanding of most of what's going on here. It does seem a bit at odds with the comment here https://github.com/nick8325/quickcheck/blob/7ff70fe9da61c95180ec90bd1ff074597efce37d/src/Test/QuickCheck/Test.hs#L228, which would make me expect `[0,1,2,...`.

MaximilianAlgehed:
Fixes #338

### DIFF
--- a/src/Test/QuickCheck/Test.hs
+++ b/src/Test/QuickCheck/Test.hs
@@ -388,7 +388,7 @@ runATest st f =
               -- Don't add coverage info from this test
               st{ numDiscardedTests         = numDiscardedTests st' + 1
                 , numRecentlyDiscardedTests = numRecentlyDiscardedTests st' + 1
-                , maxSuccessTests = fromMaybe (maxSuccessTests st) (maybeNumTests res)
+                , maxSuccessTests           = fromMaybe (maxSuccessTests st) (maybeNumTests res)
                 , maxDiscardedRatio         = fromMaybe (maxDiscardedRatio st) (maybeDiscardedRatio res)
                 , randomSeed                = rnd2
                 } f
--- a/src/Test/QuickCheck/Test.hs
+++ b/src/Test/QuickCheck/Test.hs
@@ -206,9 +206,8 @@ withState a test = (if chatty a then withStdioTerminal else withNullTerminal) $ 
                  , maxSuccessTests           = maxSuccess a
                  , coverageConfidence        = Nothing
                  , maxDiscardedRatio         = maxDiscardRatio a
-                 , computeSize               = case replay a of
-                                                 Nothing    -> computeSize'
-                                                 Just (_,s) -> computeSize' `at0` s
+                 , replayStartSize           = snd <$> replay a
+                 , maxTestSize               = maxSize a
                  , numTotMaxShrinks          = maxShrinks a
                  , numSuccessTests           = 0
                  , numDiscardedTests         = 0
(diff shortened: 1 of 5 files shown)


MaximilianAlgehed:
@rewbert thoughts?

Rewbert:
This whole discard thing is such a thorn in my side. The choice to compute sizes in this way seems to have been arbitrarily chosen initially (Koen looked guilty when I asked him about this!), so in my mind this is "equally arbitrary".

I have thought a lot about this, and have even tried out a few different approaches altogether, but have not landed in anything that is obviously superior. I think the question of sizes might be a more complicated one than it appears initially.

Regarding the eventual future merge of the parallel branch, this will not impact me negatively.

src/Test/QuickCheck/Test.hs
```diff
@@ -228,9 +228,16 @@ withState a test = (if chatty a then withStdioTerminal else withNullTerminal) $
...
           | otherwise =
-            ((n `mod` maxSize a) * maxSize a `div` (maxSuccess a `mod` maxSize a) + d `div` 10) `min` maxSize a
+            ((n `mod` maxSize a) * maxSize a `div` (maxSuccess a `mod` maxSize a) + d `div` dDenom) `min` maxSize a
+        -- The inverse of the rate at which we increase size as a function of discarded tests
+        -- if the discard ratio is high we can afford this to be slow, but if the discard ratio
+        -- is low we risk bowing out too early
+        dDenom
+          | maxDiscardRatio a > 0 = (maxSuccess a * maxDiscardRatio a `div` 3) `clamp` (1, 10)
+          | otherwise = 1 -- Doesn't matter because there will be no discards allowed
+        clamp x (l, h) = min l (max x h)
```
UlfNorell, line 240:
```suggestion
        clamp x (l, h) = max l (min x h)
```

tests/DiscardRatio.hs
```diff
@@ -40,3 +40,6 @@ main = do
   quickCheckNo $ withDiscardRatio 1 p50
   putStrLn "Expecting success (discard ratio 40): x < 50 ==> True"
   quickCheckYes $ withDiscardRatio 40 p50
+
+  let p k k' = k /= k' ==> (k :: Int) /= k'
+  quickCheckYes $ withMaxSuccess 1 p
```
UlfNorell, line 45: Maybe add a comment referencing #338

UlfNorell (changes requested):

src/Test/QuickCheck/Test.hs
MaximilianAlgehed: 🤦 50% probability of getting it 100% wrong 50% of the time...

UlfNorell (approved):

MaximilianAlgehed:
For posterity, I had to factor out `computeSize` from `State` because the `computeSize` definition was set too early based on the `Args` arguments to `quickCheckWith` and thereby would ignore `withMaxSuccess` and `withDiscardRatio`.

PR #374, opened 2024-03-21
introduce `suchThatDiscard`

### LINKED ISSUE #363: discard is too lazy
I expect using `discard` in `Gen` to discard the current case, but it doesn't discard unless I force it:
```hs
-- unexpectedly generates values
sample (discard >>= \_ -> arbitrary :: Gen Int)

-- correctly errors
sample (discard >>= \() -> arbitrary :: Gen Int)
```

MaximilianAlgehed:
Closes #363

The tricky thing here is the lazy behaviour of `Gen` and `forAll`. That's what makes me unsure whether or not `suchThatDiscard` is a good idea in practice.

### DIFF
--- a/src/Test/QuickCheck/Gen.hs
+++ b/src/Test/QuickCheck/Gen.hs
@@ -35,6 +35,7 @@ import Control.Applicative
   ( Applicative(..) )
 
 import Test.QuickCheck.Random
+import Test.QuickCheck.Exception
 import Data.List (sortBy)
 import Data.Ord
 import Data.Maybe
@@ -292,6 +293,16 @@ gen `suchThatMaybe` p = sized (\n -> try n (2*n))
     | otherwise = do
         x <- resize m gen
         if p x then return (Just x) else try (m+1) n
+
+-- | Tries to generate a value that satisfies a predicate.
+-- If it fails to do so it discards the test case if the result
+-- is used in the test.
+suchThatDiscard :: Gen a -> (a -> Bool) -> Gen a
+suchThatDiscard g p = do
+  a <- g
+  if p a
+  then pure a
+  else discard
 
 -- | Randomly uses one of the given generators. The input list
 -- must be non-empty.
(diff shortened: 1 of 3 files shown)


src/Test/QuickCheck/Gen.hs
```diff
@@ -293,6 +294,16 @@ gen `suchThatMaybe` p = sized (\n -> try n (2*n))
         x <- resize m gen
         if p x then return (Just x) else try (m+1) n
 
+-- | Tries to generate a value that satisfies a predicate.
+-- If it fails to do so it discards the test case if the result
+-- is used in the test.
+suchThatDiscard :: Gen a -> (a -> Bool) -> Gen a
+suchThatDiscard g p = do
+  a <- g
```
MaximilianAlgehed, line 302: Another option here is `(\ a -> if p a then a else discard) <$> g` - which I suppose is _slightly_ more lazy than this alternative?

nick8325:
What's the use case for this?

It's unfortunate that this has a similar name to `suchThat` and `suchThatMaybe` but behaves rather differently - in that it only tries the generator once but they retry until they succeed.

MaximilianAlgehed:
I should clarify. I don't feel particularly strongly about this idea (and the name does suck, perhaps `discardUnless` would be better?) but I was hoping it might provide a nicer interface to discard for #363 but that's just an idea.

MaximilianAlgehed:
I'll clarify some more. The problem raised by the issue is basically that `unless (p a) discard >> pure a` is a pitfall. Its clear what you intend but that is not what you get. By introducing a `suchThat` like thing you can express this intention in a natural way that does what you want.

We can have a separate conversation about what the role of `discard` is and whether or not its a good idea to use it in `Gen` but the problem raised by the issue is that since its there people are using it.

UlfNorell (changes requested):
I think `discardUnless` is a better name

UlfNorell (approved):

nick8325:
I must admit, I'm a bit sceptical that people will actually use this, just because I always find the `suchThat` API a bit clunky to use. But I've no objection to merging it.

PR #379, opened 2024-03-23
Documentation warning about the instances for `Arbitrary (Set a)` and friends being over the abstract set

MaximilianAlgehed:
Comments welcome


MaximilianAlgehed:
comments welcome from @treeowl and @phadej

UlfNorell (approved):

treeowl:
I would say more about the consequences (i.e., who should worry about this). Basically, only people digging into `Data.Set.Internal` need to care.

MaximilianAlgehed:
> I would say more about the consequences (i.e., who should worry about this). Basically, only people digging into `Data.Set.Internal` need to care.

Good idea, thank you!

PR #380, opened 2024-03-26
For shrinks, (maxBound :: Int) is too many

stephen-smith:
This PR is for master, but the patch applies for any version after 2.14.1.

I have a separate branch with the same change going back to 2.9.x

Prompted by https://recurse.social/@shapr/112141556001579384 which isn't a proper issue, but does show how this is affecting users recently.

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -1,6 +1,6 @@
 Name: QuickCheck
 Version: 2.14.1
-Cabal-Version: >= 1.8
+Cabal-Version: >= 1.10
 Build-type: Simple
 License: BSD3
 License-file: LICENSE
@@ -70,6 +70,7 @@ flag old-random
 library
   Hs-source-dirs: src
   Build-depends: base >=4.3 && <5, containers
+  Default-language: Haskell2010
 
   -- New vs old random.
   if flag(old-random)
@@ -157,7 +158,7 @@ library
 
   -- LANGUAGE pragmas don't have any effect in Hugs.
   if impl(hugs)
-    Extensions: CPP
+    Default-Extensions: CPP
 
   if impl(uhc)
     -- Cabal under UHC needs pointing out all the dependencies of the
@@ -168,6 +169,7 @@ library
 
 Test-Suite test-quickcheck
     type: exitcode-stdio-1.0
+    Default-language: Haskell2010
     hs-source-dirs:
         examples
     main-is: Heap.hs
@@ -177,6 +179,7 @@ Test-Suite test-quickcheck
 
 Test-Suite test-quickcheck-gcoarbitrary
     type: exitcode-stdio-1.0
+    Default-language: Haskell2010
     hs-source-dirs: tests
     main-is: GCoArbitraryExample.hs
     build-depends: base, QuickCheck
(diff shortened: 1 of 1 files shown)


MaximilianAlgehed:
I can think of plenty of valid use cases for shrinking more than 10^4 times in a normal property. From the thread you link it appears the problem is the broken shrinker - not the liberal behaviour of `quickCheck`? Problems with shrinking can always be mitigated by setting `maxShrinks` in `Args`.

The difficult question here is what the right tradeoff between protecting against shrink loops and making sure the default behaviour gives you small counterexamples is. I lean towards latter because it's my hypothesis that most users either (1) won't make the shrink loop mistake or more importantly (2) are mostly using the built-in `Arbitrary` instances.

I don't think this is a matter of bike-shedding.

Rewbert:
This code base already contains several magic numbers, and introducing more might not be a good idea. Furthermore, I am right now in a situation where my shrinker might require thousands of steps before it is done, so this could become a problem.

Letting the user overwrite this themselves rather than setting such a default seems more appropriate. The problem you are referencing was discovered to be an issue in the shrinker itself, not QuickCheck.

nick8325:
When I try this out in GHCi, it's pretty obvious that shrinking is the problem, because QuickCheck prints the number of shrink steps tried so far, like so:

```
ghci> quickCheck prop_shrinks_forever
*** Failed! Falsified (after 1 test and 5352847 shrinks)...
```

What you see on the screen is the number of shrinks increasing endlessly.

So I guess the problem here is that whatever test runner was being used doesn't report this information? Can we find a way to fix that instead?

stephen-smith:
> Problems with shrinking can always be mitigated by setting maxShrinks in Args.

> Letting the user overwrite this themselves rather than setting such a default seems more appropriate.

I disagree that allowing user to specify a value means the software does not need to contain a sane default.

> more than 10^4 times

Then choose something else, but `maxBound :: Int` is > 10^18, several orders of magnitude too large than is remotely reasonable.

> The problem you are referencing was discovered to be an issue in the shrinker itself

> I guess the problem here is that whatever test runner was being used doesn't report this information

I never claimed the shrinker shouldn't be fixed.  Choosing a sane default is largely a separate issue.  *No one* is going to wait through 2^63 shrinks on _any_ test case, it would simply take too long.

The default should be _significantly_ lower than `maxBound :: Int`.

MaximilianAlgehed:
My preferred default would be infinity - not because I expect anyone to wait for infinity but because any other default would be arbitrary and if it isn't functionally infinity it would get in the way of some people while giving confusingly non-minimal minimal counterexamples to others. Therefore, my vote is in favour of functionally infinity, i.e. `maxBound :: Int`.

Rewbert:
I agree. There should not be a default (which maxBound, in practice, achieves). If you need one, you can always overwrite it yourself.

UlfNorell:
An important property of shrunk counterexamples is that they cannot be shrunk any further. This can be very helpful when figuring out why a property fails. You lose this property if you hit the `maxShrinks` limit.

shapr:
If shrinking progress were displayed, that would have solved my problem.
I'll see if the test frameworks I use can display shrinking as it happens.

MaximilianAlgehed:
I'm going to close this as wontfix. If anyone feels like they have the energy making an issue detailing the pitfalls around shrink loops and suggesting some lightweight ways of getting around it I would be happy to engage there and see if we can work something out that helps people avoid this pitfall in the future.

PR #383, opened 2024-03-26
Fix the interaction between `checkCoverage` and `maxSuccess` and `discard`

### LINKED ISSUE #382: `checkCoverage` interacts poorly with discards
The following has been a thorn in my side in another project and here is a minimal failing example:
```haskell
prop_discardCoverage :: Property
prop_discardCoverage = forAll (sized $ \ n -> pure n) $ \ x -> cover 10 True "label" $ x /= 99 ==> True
```
What happens is that we get a failure like:
```
$> quickCheck $ checkCoverage prop_discardCoverage 
*** Gave up! Passed only 99 tests; 1 discarded tests (100% label).
```
when we expect the property to succeed. The reason for this is that on the 99th test (with size 98) we run `addCoverageCheck` which adds a `once` around the property. In the next test we are "unlucky" (in the real world this test is also flaky!!) and hit the discarded test, causing the whole test to be discarded because we discarded a `once`.

I have a fix for this in the pipeline.

MaximilianAlgehed:
Previously `withMaxSuccess` was discarded by `checkCoverage`. This is unfortunate because there are legitimate use cases for having both turned on at the same time. This PR fixes that. Furthermore we fix an unfortunate interaction between `discard` and `checkCoverage`.

Closes #382

### DIFF
--- a/src/Test/QuickCheck/Test.hs
+++ b/src/Test/QuickCheck/Test.hs
@@ -227,6 +227,8 @@ withState a test = (if chatty a then withStdioTerminal else withNullTerminal) $ 
 
 computeSize :: State -> Int
 computeSize MkState{replayStartSize = Just s,numSuccessTests = 0,numRecentlyDiscardedTests=0} = s
+-- NOTE: Beware that changing this means you also have to change `prop_discardCoverage` as that currently relies
+-- on the sequence produced by this function.
 computeSize MkState{maxSuccessTests = ms, maxTestSize = mts, maxDiscardedRatio = md,numSuccessTests=n,numRecentlyDiscardedTests=d}
     -- e.g. with maxSuccess = 250, maxSize = 100, goes like this:
     -- 0, 1, 2, ..., 99, 0, 1, 2, ..., 99, 0, 2, 4, ..., 98.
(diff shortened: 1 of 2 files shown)


src/Test/QuickCheck/Test.hs
```diff
@@ -343,30 +397,30 @@ showTestCount st =
...
+         updateState st0 = addNewOptions $ st0{ randomSeed = rnd2 }
 
-         addNewOptions st0 = st0{ maxSuccessTests = fromMaybe (maxSuccessTests st0) (maybeNumTests res)
+         addNewOptions st0 = st0{ maxSuccessTests   = fromMaybe (maxSuccessTests st0) (maybeNumTests res)
                                 , maxDiscardedRatio = fromMaybe (maxDiscardedRatio st0) (maybeDiscardedRatio res)
-                                , numTotMaxShrinks = fromMaybe (numTotMaxShrinks st0) (maybeMaxShrinks res)
-                                , maxTestSize = fromMaybe (maxTestSize st0) (maybeMaxTestSize res)
+                                , numTotMaxShrinks  = fromMaybe (numTotMaxShrinks st0) (maybeMaxShrinks res)
+                                , maxTestSize       = fromMaybe (maxTestSize st0) (maybeMaxTestSize res)
+                                , expected          = expect res
```
UlfNorell, line 423: We now set `expected` also in the discard case, but I don't think it makes any difference..?

UlfNorell (approved):

src/Test/QuickCheck/Test.hs
MaximilianAlgehed: It shouldn't.

PR #384, opened 2024-03-26
Make Arbitrary Map instances strict

neilmayhew:
Strict and lazy Maps share the same type so it's not possible to have separate instances for them. The existing instance uses the lazy interface and thus produces a map containing lazy values even when the client is using the strict interface. This is a violation of the invariant of a strict Map, whereas there's no invariant to be violated with a lazy Map. It seems better, therefore, to make the instance default to producing strict values.

Note that strict maps evaluate their values only to WHNF, so a `Map Int (Maybe Char)` would still contain thunks.

If introducing a dependency on `nothunks` (even if only for testing) might be a problem, the commit containing the test can just be dropped. I added it mainly as a way of proving to myself that the fix works. (The test fails without the commit containing the fix.)

I also didn't add any conditional compilation for non-GHC builds, which I assume runs against the traditional philosophy of QuickCheck. If you want to preserve this, that would be another reason to drop the testing commit.

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -249,6 +249,15 @@ Test-Suite test-quickcheck-split
     main-is: Split.hs
     build-depends: base, QuickCheck
 
+Test-Suite test-quickcheck-strictness
+    type: exitcode-stdio-1.0
+    Default-language: Haskell2010
+    hs-source-dirs: tests
+    main-is: Strictness.hs
+    build-depends: base, QuickCheck, containers
+    if !flag(templateHaskell) || !impl(ghc >= 7.10) || impl(haste)
+        buildable: False
+
 Test-Suite test-quickcheck-misc
     type: exitcode-stdio-1.0
     Default-language: Haskell2010
--- a/src/Test/QuickCheck/Arbitrary.hs
+++ b/src/Test/QuickCheck/Arbitrary.hs
@@ -165,9 +165,14 @@ import GHC.Generics
 #endif
 
 import qualified Data.Set as Set
-import qualified Data.Map as Map
 import qualified Data.IntSet as IntSet
+#if MIN_VERSION_containers(0,5,0)
+import qualified Data.Map.Strict as Map
+import qualified Data.IntMap.Strict as IntMap
+#else
+import qualified Data.Map as Map
 import qualified Data.IntMap as IntMap
+#endif
 import qualified Data.Sequence as Sequence
 import qualified Data.Tree as Tree
 import Data.Bits
--- a/tests/Strictness.hs
+++ b/tests/Strictness.hs
@@ -1,0 +1,42 @@
+-- Strictness tests.
+
+{-# LANGUAGE CPP, TemplateHaskell #-}
+import Test.QuickCheck
+
+import Control.Exception (Exception (..), throw)
+
+#if MIN_VERSION_containers(0,5,0)
+import Control.Exception (evaluate, try)
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+#endif
+
+data Thunk
+
+instance Arbitrary Thunk where
+  arbitrary = throw ThunkError
+
+instance Show Thunk where
+  show _ = "Thunk"
+
+data ThunkError = ThunkError deriving Show
+
+instance Exception ThunkError
+
+prop_strictMap :: Property
+#if MIN_VERSION_containers(0,5,0)
+prop_strictMap = again . ioProperty $ do
+  m <- generate arbitrary
+  result <- try $ evaluate m :: IO (Either ThunkError (Map Int Thunk))
+  pure $ case result of
+    Right _ | not (Map.null m) -> counterexample ("Thunks in Map: " ++ show m) False
+    _ -> property True
+#else
+prop_strictMap = once $ property True
+#endif
+
+return []
+main :: IO ()
+main = do
+  True <- $quickCheckAll
+  return ()


QuickCheck.cabal
```diff
@@ -254,7 +254,7 @@ Test-Suite test-quickcheck-misc
     Default-language: Haskell2010
     hs-source-dirs: tests
     main-is: Misc.hs
-    build-depends: base, QuickCheck
+    build-depends: base, QuickCheck, containers, nothunks
```
phadej, line 257: I'd say `nothunks` is a bad idea, as it doesn't support all the GHCs `QuickCheck` does.

MaximilianAlgehed, line 257: I don't mind having the test at all, that's a good thing. But perhaps we could have this test in its own tiny test suite? That way it will be easier to turn off without turning everything else off (if need be).

phadej: flag for tests is a ugly. Users don't care about it, but they see it. Also you'll forget to toggle it.

Having a module which is conditional on `if impl(ghc >= ...)` is better.

MaximilianAlgehed: yes, derp. that's what I meant.

MaximilianAlgehed:
This is a tricky thing. On the one hand we don't want the behaviour of `QuickCheck` to be different across versions, on the other hand breaking invariants in `containers` seems like a big no-no. I'm leaning towards putting the import behind a `#if MIN_VERSION_containers(0,5,0)`. I can't think of a legitimate use case for making the `Arbitrary` instance incompatible with one of  the main interfaces for `Data.Map`.

It is unfortunate, for QuickCheck's purposes, that `containers` is designed the way it is with no difference in the type to distinguish the variants with and without an invariant. But that is a totally legitimate design, with good reason, that we can't ignore.

If that's very objectionable I would go for having a `Strict` modifier in the style of `Postive` etc that depends on `DeepSeq`. In fact, having the `Strict` modifier would be a good idea _regardless_ of this change...

neilmayhew:
> If that's very objectionable I would go for having a `Strict` modifier in the style of `Positive` etc that depends on `DeepSeq`.

It would be very time-consuming for us to go through every every one of our many `Arbitrary` instances that uses a strict `Map` and add the modifier. So I would prefer to avoid that if at all possible.

> In fact, having the `Strict` modifier would be a good idea _regardless_ of this change...

Agreed!

nick8325:
If we want to avoid importing `Data.Map.Strict`, we could just use `seq` on the values in the map before doing `Map.fromList`:

```
fromList' :: Ord k => [(k, v)] -> Map.Map k v
fromList' xs =
  foldr seq () (map snd xs) `seq` Map.fromList xs
```

neilmayhew:
> If we want to avoid importing `Data.Map.Strict`, we could just use `seq` on the values in the map before doing `Map.fromList`:

That's a nice solution. However, I feel it's better to let `Data.Map.Strict` be responsible for enforcing its own invariant rather than having QuickCheck guess at it. We can do that with the `#if MIN_VERSION_containers(0,5,0)` that Max suggested, and we already have precedent for that in `src/Test/QuickCheck/Test.hs`.

neilmayhew:
> If I can figure out a way of testing for thunks without using nothunks I'll submit another PR.

I have what I think is a nice solution, and I'll include it with this PR. I've moved it to a separate module, and I've used the same CPP conditional to disable the test when appropriate.

neilmayhew:
CI here needs approval, but it [succeeded](https://github.com/neilmayhew/quickcheck/actions/runs/8443694510) in my fork

src/Test/QuickCheck/Arbitrary.hs
```diff
@@ -165,9 +165,19 @@ import GHC.Generics
...
+import qualified Data.Map.Strict as Map
+import qualified Data.IntMap.Strict as IntMap
+#else
+import qualified Data.Map as Map
+import qualified Data.IntMap as IntMap
+#endif
+#else
+import qualified Data.Map as Map
 import qualified Data.IntMap as IntMap
+#endif
```
MaximilianAlgehed, line 180: This seems more complicated than it needs to be given that the old code was just straight-up imports?

tests/Strictness.hs
```diff
@@ -0,0 +1,42 @@
+-- Strictness tests.
+
+{-# LANGUAGE CPP, TemplateHaskell #-}
+import Test.QuickCheck
+
+import Control.Exception (Exception (..), throw)
+
+#if MIN_VERSION_containers(0,5,0)
```
MaximilianAlgehed, line 8: I'm inclined to say we don't need these if we have the right GHC version bound in the `cabal` file.

MaximilianAlgehed (approved):

src/Test/QuickCheck/Arbitrary.hs
neilmayhew: I agree, but since this is in a module that will always be included in the build I think it has to be done with a CPP conditional.

Regarding the `#if defined`, this is the pattern currently being used in `src/Test/QuickCheck/Test.hs` and I assumed I should follow it. However, if we can assume that the build is always going to be done using Cabal, the macro will always be defined so we can drop this extra complexity.

tests/Strictness.hs
neilmayhew: You mean, don't build the module at all? We could do that, but it means using the GHC version as a conditional rather than the more precise and meaningful conditional on the version of `containers`. It's also guaranteed to match the conditional on the feature that's being tested.

It would be possible to add a Cabal flag that the solver can set automatically via a bound on `containers`, but I feel it's better not to add to the complexity of the Cabal file.

src/Test/QuickCheck/Arbitrary.hs
neilmayhew: I removed the defensive `#if defined`

MaximilianAlgehed:
I think this is good to go as-is. Unless there are any reservations from anyone?

UlfNorell (approved):
Looks good. I can imagine there might be cases where you want the lazy behaviour, but in that case you can easily wrap the values to recover it.

tests/Strictness.hs
MaximilianAlgehed: That sounds like too much work for too little gain. Let's keep it like this.

src/Test/QuickCheck/Arbitrary.hs
```diff
@@ -165,9 +165,9 @@ import GHC.Generics
 #endif
 
 import qualified Data.Set as Set
-import qualified Data.Map as Map
+import qualified Data.Map.Strict as Map
```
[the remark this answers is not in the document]
MaximilianAlgehed: Fixed by the check version pragma

PR #385, opened 2024-03-27
Silence warning for partial functions on ghc >=9.8

UlfNorell:


MaximilianAlgehed:
What's the case against just removing said partial function calls from the code base?

UlfNorell:
These are the partial functions we get warnings for:

- https://github.com/nick8325/quickcheck/blob/97ee742f1a17c5614812916e78117e419be8a4a1/src/Test/QuickCheck/Arbitrary.hs#L1209
Could be `drop 1`.

- https://github.com/nick8325/quickcheck/blob/97ee742f1a17c5614812916e78117e419be8a4a1/src/Test/QuickCheck/Function.hs#L143
`head (someList ++ [d])` is a useful idiom. Not sure I want to change this to `maybe fst d (uncons someList)`... (when was `uncons` added to `Data.List`?). I suppose we could define `headWithDefault`.

- https://github.com/nick8325/quickcheck/blob/97ee742f1a17c5614812916e78117e419be8a4a1/tests/Split.hs#L16-L17
`map head . group` is another common idiom that I'm not sure I want to rewrite just to avoid the warning.

MaximilianAlgehed (approved):

phadej:
FWIW, `map head . group` is exactly that if you use `head` and `group` from [`Data.List.NonEmpty`](https://hackage.haskell.org/package/base-4.19.1.0/docs/Data-List-NonEmpty.html)

PR #387, opened 2024-03-27
Consider labels not covered at all insufficiently covered

### LINKED ISSUE #254: `coverTable` is silent if key isn't `tabulate`d
When used as intended, `coverTable` gives a coverage warning:
```
> quickCheck $ coverTable "A" [("B", 10)] $ tabulate "A" ["A"] ()
+++ OK, passed 1 test.

A (1 in total):
100% A

Table 'A' had only 0% B, but expected 10%
```

If one messes up and accidentally uses the wrong key, say "C" instead of "A", when calling `coverTable` it silently succeeds:

```
> quickCheck $ coverTable "C" [("B", 10)] $ tabulate "A" ["A"] ()
+++ OK, passed 1 test.

A (1 in total):
100% A
```

It would be nice to have a warning in this case?

(In my actual use case I abuse the keys of `tabulate` to be from-states and the values to be to-states in a finite state automata.)

MaximilianAlgehed:
Closes #254

### DIFF
--- a/src/Test/QuickCheck/Test.hs
+++ b/src/Test/QuickCheck/Test.hs
@@ -503,10 +503,21 @@ labelsAndTables st = (theLabels, theTables)
       paragraphs $
         [ showTable (sum (Map.elems m)) (Just table) m
         | (table, m) <- Map.toList (S.tables st) ] ++
-        [[ (case mtable of Nothing -> "Only "; Just table -> "Table '" ++ table ++ "' had only ")
-         ++ lpercent n tot ++ " " ++ label ++ ", but expected " ++ lpercentage p tot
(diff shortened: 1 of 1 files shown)


MaximilianAlgehed:
My concern with this approach to fixing the linked issue is that it introduces a strange interaction that I'm not necessarily a fan of. If you've had 0 instances of something after 100 tests and you're running `checkCoverage` we will now reject the test - even if your required coverage was very low (meaning that before QuickCheck would have run more tests). I'm not 100% convinced this is a problem (you shouldn't be requiring 2% coverage on something!) but we can maybe re-work this PR to make that case work out fine?

MaximilianAlgehed:
This was based on an incorrect understanding of `coverTable`. Closing this and the linked issue.

PR #367, opened 2024-03-18
Add `withDiscardRatio` with a similar interface to `withMaxSuccess`

MaximilianAlgehed:
Also clean up a little bit.

PR #391, opened 2024-04-03
Version bump to 2.15

MaximilianAlgehed:
Thoughts welcome, are there any outstanding PRs that we should try to incorporate in this release? I would be happy if we could get #387 and maybe #376 (but I'm not sure if the design work is done there yet).

### DIFF
--- a/QuickCheck.cabal
+++ b/QuickCheck.cabal
@@ -1,5 +1,5 @@
 Name: QuickCheck
-Version: 2.14.3
+Version: 2.15.0
 Cabal-Version: >= 1.10
 Build-type: Simple
 License: BSD3
--- a/changelog
+++ b/changelog
@@ -1,3 +1,33 @@
+QuickCheck 2.15
+	* BREAKING: properties without quantification are no longer implicitly
+	  wrapped in once
+	* Add witness property modifier and witnesses field in result on
+	compilers that support Typeable to allow conveniently exposing
+	counterexamples
+	* Add assertException, assertExceptionIO, assertDeepException, and
+	assertDeepExceptionIO (thanks to Alexey Kuleshevich)
+	* Add withMaxSize, withMaxShrinks, and withMaxDiscardRatio combinators
+	* Add recheck
+	* Add HasCallStack to partial top-level functions
+	* Updated some documentation
+	* Fix overflow bug when calculating discard ratio
+	* Fix error printing bug when shrinks are discarded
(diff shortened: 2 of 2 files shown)


Rewbert:
I think both of the mentioned PRs could be included, no? #387 seems like low-hanging fruit, and #367 does not seem to affect existing users negatively. They can opt in to use the new interface if they want to, but retain the old behavior if they don't take any action.

MaximilianAlgehed:
@Rewbert Whoops, I meant #376 not #367

MaximilianAlgehed:
Also, c.f. my comment over on #387 for why I have some issues there that we could, in principle, ignore but which may come back to bite us in the future.

Rewbert:
Right, that PR is a bit more substantial. It is a feature that makes a lot of sense, but as you say, there is more design-work involved there than the other PRs. I like it, though!

What I've done in the past is to install a monitor that fires `whenFail`, which writes the counterexample to an `MVar`. A bit crude.

MaximilianAlgehed:
There is also @nick8325's package that modifies the interface to QuickCheck that solves this in another hacky-but-elegant way:
https://hackage.haskell.org/package/quickcheck-with-counterexamples-1.2/candidate/docs/Test-QuickCheck-Counterexamples.html

Rewbert:
I just looked at it, I like it as well! Your implementation is smaller and perhaps 'easier' to retrofit onto QuickCheck (I did not look at Nicks code yet, only the API), whereas Nicks might give users a bit more help with accessing the counterexamples. Maybe there is a third alternative that takes the best from both... I shall look at Nicks code and wrap my head around it.

MaximilianAlgehed:
It would be nice to have #393 merged too in the next release

nick8325:
Looks like a good release! #376 is the main one to be decided on still, right?

One change is missing from the changelog: properties are no longer implicitly wrapped in `once` (c606b7b3). I guess that's a breaking change (e.g. properties of type `prop_whatever :: IO Bool` will now be tested 100 times instead of once), so maybe according to PVP we should bump the version to 2.15?

MaximilianAlgehed:
@nick8325 I've updated the changelog to reflect the breaking `once` change, updated the version bound, and I've added a changelog entry for the now-completed #376.

There was more discussion in #376 about more aggressive changes to QuickCheck to make it more like quickcheck-counterexample but I think that can wait until a later release and until we've had time to take the current changes for a spin.

PR #404, opened 2024-05-21
Resolve Haddock out-of-scope identifier warnings

pgujjula:
Fix all broken Haddock links. Now Haddock doesn't emit any warnings about out-of-scope identifiers. Warnings about unknown link destinations more complicated to resolve and are not handled in this PR.

### DIFF
--- a/src/Test/QuickCheck.hs
+++ b/src/Test/QuickCheck.hs
@@ -50,8 +50,8 @@ them to:
 +++ Ok, passed 100 tests.
 @
 
-That's because GHCi will default any type variables in your property to '()', so in the example
-above @quickCheck@ was really testing that '()' is equal to itself. To avoid this behaviour it
+That's because GHCi will default any type variables in your property to @()@, so in the example
+above @quickCheck@ was really testing that @()@ is equal to itself. To avoid this behaviour it
 is best practise to monomorphise your polymorphic properties when testing:
 
 @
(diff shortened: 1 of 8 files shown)


src/Test/QuickCheck.hs
```diff
@@ -50,8 +50,8 @@ them to:
 +++ Ok, passed 100 tests.
 @
 
-That's because GHCi will default any type variables in your property to '()', so in the example
-above @quickCheck@ was really testing that '()' is equal to itself. To avoid this behaviour it
+That's because GHCi will default any type variables in your property to @()@, so in the example
+above @quickCheck@ was really testing that @()@ is equal to itself. To avoid this behaviour it
```
pgujjula, line 54: `()` is not actually in the Haddocks for Prelude, so linking to `()` would require pulling in `ghc-prim` as a dependency. So we just make `()` monospaced instead of linking to it.

MaximilianAlgehed:
Would you mind rebasing on `master` and we can try to get this over the line ASAP.

pgujjula:
Hey @MaximilianAlgehed, sorry I missed your comment, I just rebased against the latest master and added one more fix that for a warning that appeared in the last few months related to `isSuccess` here: [(link)](https://github.com/nick8325/quickcheck/pull/404/files#diff-c82977f0bb68f33cee8079df52b389b4060f0336e5127b2a289ddbfdc3ebc3f9L22).

cc @UlfNorell

MaximilianAlgehed (approved):

Issue #352, opened 2023-03-30
Printing is slow!


Rewbert:
Printing is super expensive; I just ran a property where each test case takes very little time. 200 000 tests took 14 seconds, but 200 000 tests with `chatty = False` took 4 seconds. Such a huge penalty for just printing!

`chatty = False` is not always ideal, as I am interested in the current progress of my property. There should be some half-way option where I can ask for output, but less. I am just not interested in unperceived updates :) I have a patch ready to go for this, but I document this here so that this is not forgotten.

Rewbert:
For slow tests the current implementation is fine (execution time is not dominated by printing), but for fast ones we need an FPS limiter! :)

Rewbert:
In the parallel fork of QC, this is not a problem. The final design might change, but _something_ will be done about this in 2.16.

PR #410, opened 2024-06-18
Multicore support

### LINKED ISSUE #180: Run tests on multiple cores
There are packages like [pqc](http://hackage.haskell.org/package/pqc) and [tasty-quickcheck](http://hackage.haskell.org/package/tasty-quickcheck) that allow to run tests in parallel on multiple cores. Since they run tests on different properties in parallel, they do no longer show the increasing test counter. Would it be possible to run different tests on the same property in parallel? This would still enable the live counter. Maybe it would even save some memory if tests on the same property share some data.
Maybe the parallel testing could be enabled by a new field of `Args` for the number of parallel testing threads?

Rewbert:
Support for running tests in parallel, fixes #180, and addresses #352.

Packages like Tasty already have support for running multiple properties in parallel (calling `quickCheck` more than once in parallel), whereas this work distributes all the tests concerning a _single_ property over the available cores. This is more helpful when you as a developer are working on something, as opposed to when tests are being executed in CI.

I have used things only found in base to not increase the already large list of dependencies for QC. Primarily `Control.Concurrent` and `Control.Exception`

I wrote up some more text regarding this work here: https://www.krook.dev/posts/quickercheck/quickercheck.html
The paper presenting the work was published at IFL 23', and can be found here: https://www.krook.dev/papers/quickercheck.pdf Please give it a read to get a feeling for how the implementation works. There are no details like code, but it explains in broad terms how the implementation works.

I expect it will be a nightmare to merge this (and make sure related packages like Tasty and HSpec don't break), but it will be worth it in the end. The change is quite large with some 1500 LOC added and 500 LOC removed, and so far I am the only one that's really looked at it. I've had meetings with Nick and discussed some design alternatives, but not really looked at the code. It is good if more people than me look at it. Below follows a non-exhaustive list of things that needs to be addressed, aside from the obvious crap that makes this not mergeable at the moment

- [ ] Features.hs is commented out at the moment, and needs to be fixed

- [ ] The shrinker uses the UserInterrupt exception at the moment but should use a QC internal exception instead

- [ ] Graceful combinator also uses UserInterrupt, but should use the same internal exception the shrinker uses (they must be the same)

- [ ] Should we support both offset and stride for size computations? I am not a fan of how the size is computed, but what I've done is retrofit this solution over the existing implementation.

- [ ] I moved some things around. I can't remember everything, but I think I e.g. moved `State` to another module to avoid some cyclic problems

- [ ] I added a CTRL-C handler, but I know @nick8325  doesn't like this as users might have their own handlers installed. This should go (easy to do)

- [ ] Currently, the same flag is used to indicate the number of testers as well as the number of shrinkers (the -N flag populates this field). Experimentation showed that fewer workers are better when shrinking. On my 4-core 8-thread machine, even if I could utilise 8 threads for testing, I reached an optimal when shrinking if I used 3-4 workers.

- [ ] Currently, the right to run a test is claimed before the test actually starts running. An effect of this is that if you run 100 tests and 99 of them finish very fast, whereas the 100th one is an outlier and takes a while, it will seem as if the tests are stuck at 99/100. Another possible design is that the right to run a test is claimed when you report the results, meaning that you have already run the test. An effect of this is that we won't hang like described above, but also that we will explore a different space of tests. If a slow-running test is an indication that a bug is present, we are less likely to find it.

- [ ] When I made my fork there was very little action in the repository. Since then, @MaximilianAlgehed  has managed to squash a lot of old issues! I tried to reapply some of the changes that were applied to the testing loop, but there were some that I've not reapplied yet. TODO Look up which (I have this written on a note at work) and add as points here

- [ ] I need to go over all the documentation and update it to reflect the current state of the implementation. I documented it some while back but can't quite remember if I changed relevant stuff since then.

While this MR is quite unpolished, with input from the other maintainers I can put in some more work and get this in a proper state.


README
```diff
@@ -1,3 +1,36 @@
...
+    -- tag: <commit hash of the version you want>
+```
+
+2: You need to add some flags when you compile your code. Specifically, `-threaded -feager-blackholing -rtsopts`.
+
+3: Finally, all that is left is to change the call to `quickCheck` with a call to `quickCheckPar`. If you don't want parallel shrinking, you should call `quickCheckParWith (stdArgs { parallelShrinking = False}) property`.
+
+4: You also need to pass in the runtime option that actually creates more HECs. You need to either instead of `cabal run executable` do `cabal run executable -- +RTS -N` or `-Nx` where x is a number between 1 and the number of cores you have. You can also ddd another compilation option `-with-rtsopts=-N` or `-with-rtsopts=-Nx`
+
+As a sanity check of whether you are using my fork or not, if you run `quickCheckPar` with just 1 HEC available, the word `donkey` will be printed.
```
MaximilianAlgehed, line 24: This is probably not something we want in the final readme ;)

src/Test/QuickCheck/Features.hs
```diff
@@ -17,16 +16,16 @@ features :: [String] -> Set String -> Set String
...
+-- prop_noNewFeatures :: Testable prop => Set String -> prop -> Property
+-- prop_noNewFeatures feats prop =
+--   mapResult f prop
+--   where
+--     f res =
+--       case ok res of
+--         Just True
+--           | not (features (P.labels res) (Set.fromList (P.classes res)) `Set.isSubsetOf` feats) ->
+--             res{ok = Just False, P.reason = "New feature found"}
+--         _ -> res
```
MaximilianAlgehed, line 28: How come this needs to be commented out?

src/Test/QuickCheck/Property.hs
```diff
@@ -265,31 +269,37 @@ data CallbackKind = Counterexample    -- ^ Affected by the 'verbose' combinator
...
+  , maybeCheckCoverage  :: Maybe Confidence
     -- ^ required coverage confidence
-  , labels             :: [String]
+  , maybeDiscardedRatio :: Maybe Int
+    -- ^ maximum number of discarded tests per succesful test
+  , maybeMaxShrinks     :: Maybe Int
+    -- ^ maximum number of shrinks
+  , maybeMaxTestSize    :: Maybe Int
+    -- ^ maximum test size
+  , labels              :: [String]
```
MaximilianAlgehed, line 292: I'm slightly concerned this is showing up in the diff - these are on `master`. What's going on there?

MaximilianAlgehed (changes requested):

sol:
This is how Hspec interacts with QuikCheck:

https://github.com/hspec/hspec/blob/main/hspec-core/src/Test/Hspec/Core/QuickCheck/Util.hs

https://github.com/hspec/hspec/blob/main/hspec-core/src/Test/Hspec/Core/QuickCheck.hs#L66-L105

Bodigrim:
QuickCheck-specific parts of `tasty` are all in one file: https://github.com/UnkindPartition/tasty/blob/master/quickcheck/Test/Tasty/QuickCheck.hs

Both `tasty` and, I imagine, `hspec` parallelize test execution by running individual tests in separate threads. It's likely to be unhelpful if underlying QuickCheck properties try to spawn multiple threads as well. Would it be possible to keep default singlethreaded?

Rewbert:
> QuickCheck-specific parts of `tasty` are all in one file: https://github.com/UnkindPartition/tasty/blob/master/quickcheck/Test/Tasty/QuickCheck.hs
> 
> Both `tasty` and, I imagine, `hspec` parallelize test execution by running individual tests in separate threads. It's likely to be unhelpful if underlying QuickCheck properties try to spawn multiple threads as well. Would it be possible to keep default singlethreaded?

Yes, the default behavior in this PR is singlethreaded.

src/Test/QuickCheck.hs
```diff
@@ -270,6 +320,9 @@ module Test.QuickCheck
   , withMaxSuccess
   , within
   , discardAfter
+  , withDiscardRatio
+  , withMaxSize
+  , withMaxShrinks
```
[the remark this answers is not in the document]
Rewbert: I made my fork a while ago when the repo was stale, and since then many changes happened to master. This PR requires too much work to merge, and will be closed.

I will open another PR in a while that contains only the necessary changes for multicore support, and that is actually mergeable. This will happen after we've done some other work to QC.

I sent you an email some days ago @sol  asking precisely what hooks `hspec` installs with e.g. the `aroundProperty` function. I hope to create a nicer interface to the functionality you desire.

MaximilianAlgehed:
@Rewbert I'm closing this in favour of opening a new PR later on where we try to bring your fork up-to-speed.

PR #412, opened 2024-07-09
Bump workflow versions

MaximilianAlgehed:
Trying to bring things up to date to make sure the issues we've seen in CI on #397 aren't due to old versions.


MaximilianAlgehed:
@nick8325 here is something we need to discuss. The old ci system has stopped working due to random github actions deprecations. I've updated to the latest version of `haskell-ci` and bumped the ci script. However, a consequence of this is that we can no longer support ghc versions older than 8.0.2 in ci.

Should we make the decision to discontinue ghc 7 support?

MaximilianAlgehed:
@UlfNorell thoughts on this?

MaximilianAlgehed:
Current state of things here:
1. We can support 7.10.3 since that's supported by `setup-haskell`
2. We change to `setup-haskell` as that gives us more flexibility than depending on `haskell-ci`

UlfNorell (approved):
If nothing else this greatly simplifies the CI setup. It seems to me the question here isn't should we drop ghc-7 support, but rather getting CI to run at all. If we want to keep support for older versions that's something that can be figured out later.

PR #415, opened 2024-09-21
Speed up heavy use of mapTotalResult

MaximilianAlgehed:
@nick8325 we are _very_ confused about what's going on here. `mapTotalResult` results in repeated calls to `property` on something of type `Property`. This in turn results in repeated layers of `protectProp` which causes properties that do _a lot_ of tabulate or other similar things to blow up.

Now, our intuition is that it should be fine to only do the `protectProp` _once_, when we run the property. Is this correct? What was the reason for the convoluted instance? No tests fail if you just do `property = id` and nothing else...

Thoughts very much welcome.


MaximilianAlgehed:
Isn't the solution here to only re-install the exception handler in the case where you go from something polymorphic to Property - that way we wouldn't have to re-apply the protection to something of type property?

nick8325:
That would be reasonable, but it doesn't help in this case, does it? `tabulate` is polymorphic after all...

MaximilianAlgehed:
Yes, so it would apply it the first time, but when tabulate is applied to something that is already of type property, we don't need to apply the protection again?

nick8325:
Isn't that what your patch already does, though?

MaximilianAlgehed:
I don't get it, did you run your counterexample above on this branch? Because I really cannot see how that would succeed on master but fail on this branch?

nick8325:
Yep, on master I get a slightly different output - note the `3` at the end:

```
ghci> quickCheck prop_bad
*** Failed! (after 18 tests):                                            
Exception:
  Prelude.undefined
  CallStack (from HasCallStack):
    undefined, called at Test.hs:4:14 in main:Main
3
```

What's happening is that `forAll` is implemented roughly like this:

```
forAll gen prop = MkProperty $ do
   x <- gen
   unProperty $ counterexample (show x) $ prop x
```

and `counterexample` is polymorphic in the property type, so (on master) `prop x` gets wrapped in an exception handler, which is what allows `counterexample` to do its work. On your branch there's only an exception handler at the topmost level for this example. For the counterexample to be shown there needs to be an exception handler _inside_ the call to `counterexample`.

MaximilianAlgehed:
Right! But you only need one exception handler when you do `counterexample . counterexample` right?

nick8325:
Yes that's true! But if it's `counterexample . f . counterexample` where `f` is a user-defined function, then you might need two (i.e. if f can crash). So you have to be able to distinguish these cases.

I have an idea how to do that, let me write it up...

nick8325:
Right, here goes. Note that when I talk about "bottom" here, I really mean "crashing" - I've not thought about non-terminating properties.

Call a value of type `Property` "well-behaved" if it only ever generates non-bottom rose trees. More formally, if `prop` is well-behaved then in `do { x <- unProperty prop; ...}`, x is always non-bottom.

Note that any well-behaved Property does not need protecting. (In particular, we already maintain invariants that mean that a non-bottom rose tree must produce a non-bottom result.)

We'd like to maintain the invariant "every `Property` is well-behaved" but that's too strong (since the user can always pass in `undefined :: Property`). But, if we make `Property` an abstract type (i.e. hide `MkProperty`) then we can maintain the following invariant:

* every _non-bottom_ `Property` is well-behaved.

Then we can make protecting a property a no-op as long as the property is non-bottom:

```
protectProperty :: Property -> Property
protectProperty p
  | crashes p = MkProperty (fmap protectProp (unProperty p)) -- in reality the result here is always a test case failure
  | otherwise = p
  where
     crashes :: a -> Bool
     crashes x = -- evaluate x to WHNF and return True if it throws an exception
```

The point is that applying `protectProperty` twice, the second call will be a no-op as the property will already be in WHNF.

MaximilianAlgehed:
I like that idea a lot! I can try to implement this on the branch tomorrow - and I'll try to make a failing test of your "no counterexample" property too.

I've been thinking, wouldn't it be nice to write some QuickCheck properties to check the strictness properties of QuickCheck - one could have a little DSL that compiles to `Property` for which you could predict what the counterexamples should be and when things should fail because of exceptions etc.

MaximilianAlgehed:
What I fail to understand here is why in your example the `crashes p` case doesn't just return `MkProperty bottom` because presumably `unProperty bottom = bottom` and `fmap f bottom = bottom`?

nick8325:
Hmm, in this case I think (?) it works because `fmap` is in the `Gen` monad which is lazy (`fmap f gen = MkGen (\g n -> f (unGen gen g n))`). But I could be wrong :) In any case it's probably clearer to explicitly return a failing result there.

And yes, strictness properties would be very nice!

MaximilianAlgehed:
Ah yes, in `Gen` you get `do x <- bottom; protectProp x = protectProp
bottom` or something like that...

Sincerely
Maximilian Algehed, Ph.D.

PR #417, opened 2024-11-06
Use `displayException` instead of `show`

edsko:
This is particularly useful in ghc 9.10 and up, where `displayException` shows the attached annotations, including backtraces (if any).

### DIFF
--- a/src/Test/QuickCheck/Property.hs
+++ b/src/Test/QuickCheck/Property.hs
@@ -25,6 +25,9 @@ import System.Timeout(timeout)
 #endif
 import Data.Maybe
 import Control.Applicative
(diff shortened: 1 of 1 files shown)


MaximilianAlgehed:
Thank you for this contribution! If you could fix the CI failures (in this case you probably just need to add some pragmas guarding against old compilers) I think this would be a valuable contribution.

edsko:
Have pushed what's hopefully a fix. This is the first time in my Haskell career I have had CI fail due to a test against _hugs_, feels like a rite of passage somehow :grimacing:

MaximilianAlgehed (approved):

PR #419, opened 2025-01-06
Lower memory use of `instance Arbitrary Double`/`Float`.

### LINKED ISSUE #418: `Instance Arbitrary Double` uses lots of memory with high sizes
My code has been doing `scale (^ 8)` to generate `Double`s in between ±10^16 instead of in between ±100. But at that size, memory usage is very high since QuickCheck-2.14.3.

```
$ ghci -package QuickCheck-2.14.2 +RTS -M100M -RTS
GHCi, version 9.6.6: https://www.haskell.org/ghc/  :? for help
ghci> import Test.QuickCheck
ghci> flip traverse [1..100] $ \_ -> length . show <$> generate (scale (const 100_000_000) $ arbitrary @Double)
[20,20,19,19,19,19,20,20,21,20,20,19,20,17,20,19,21,19,20,21,18,19,20,20,20,18,20,19,20,19,19,18,19,21,20,19,18,19,20,21,20,19,20,19,19,19,18,20,19,19,21,19,19,17,20,19,16,21,20,19,18,19,20,20,19,19,21,20,20,21,20,20,19,19,19,20,20,19,19,19,19,20,19,19,17,20,20,20,20,20,18,19,20,19,19,19,20,20,20,19]

$ ghci -package QuickCheck-2.14.3 +RTS -M100M -RTS
GHCi, version 9.6.6: https://www.haskell.org/ghc/  :? for help
ghci> import Test.QuickCheck
ghci> flip traverse [1..100] $ \_ -> length . show <$> generate (scale (const 100_000_000) $ arbitrary @Double)
[20,*** Exception: heap overflow
```

(But I've also seen the heap overflow with `-M4G`.)

I guess this is due to `smallDenominators` introduced in #297, which picks a number up to `size` and then uses O(n) time and memory to index into a [tree](https://en.wikipedia.org/wiki/Calkin%E2%80%93Wilf_tree) of rationals.

Some possibilities that come to mind:

* Specify that `Arbitrary` instances might not expect sizes that large, and users need to use other generators if they want that kind of thing. I don't think it would cause me big problems, but I don't love the idea. ("I want bigger numbers -> I'll use a bigger size" is pretty natural, which is how I came across this problem; and if the numbers are embedded in some other data structure, specifying another generator might be awkward.)
* In `smallDenominators`, limit the size to some relatively small max.
* It's possible to calculate the tree values without enumerating the tree. ~~I guess this would make it O(log n) time and memory (effectively O(1) because size is `Int` not `Integer`).~~ I expect this would be a lot faster and more memory efficient for large `n`. If it's too slow for small `n`, then "build the first _ tree values, calculate directly from then on" could work.

ChickenProp:
ghci> flip traverse [1..100] $ \_ -> length . show <$> generate (scale (const 100000000) $ (arbitrary :: Gen Double))

on master, this is slow and memory hungry. (I couldn't figure out how to pass RTS options to ghci using `cabal repl`, but I've seen it take over 5 minutes and 10 GB.) With this commit it's fast.

Closes #418.

### DIFF
--- a/src/Test/QuickCheck/Arbitrary.hs
+++ b/src/Test/QuickCheck/Arbitrary.hs
@@ -723,7 +723,7 @@ instance Arbitrary Float where
     where
       smallDenominators = sized $ \n -> do
         i <- chooseInt (0, n)
-        pure (fromRational (streamNth i rationalUniverse))
+        pure (fromRational (streamNth (min i 256) rationalUniverse))
 
       uniform = sized $ \n -> do
         let n' = toInteger n
@@ -749,7 +749,7 @@ instance Arbitrary Double where
     where
       smallDenominators = sized $ \n -> do
         i <- chooseInt (0, n)
-        pure (fromRational (streamNth i rationalUniverse))
+        pure (fromRational (streamNth (min i 256) rationalUniverse))
 
       uniform = sized $ \n -> do
         let n' = toInteger n
@@ -1639,8 +1639,9 @@ streamNth :: Int -> Stream a -> a
 streamNth n (x :< xs) | n <= 0    = x
                       | otherwise = streamNth (n - 1) xs
 
--- We read into this stream only with ~size argument,
--- so it's ok to have it as CAF.
+-- We read into this stream only with ~size argument, capped to 256,
+-- so it's ok to have it as CAF. (256 chosen somewhat arbitrarily, the
+-- point is just to stop this blowing up.)
 --
 rationalUniverse :: Stream Rational
 rationalUniverse = 0 :< 1 :< (-1) :< go leftSideStream


src/Test/QuickCheck/Arbitrary.hs
```diff
@@ -723,7 +723,7 @@ instance Arbitrary Float where
     where
       smallDenominators = sized $ \n -> do
         i <- chooseInt (0, n)
-        pure (fromRational (streamNth i rationalUniverse))
+        pure (fromRational (streamNth (min i 256) rationalUniverse))
```
MaximilianAlgehed, line 726: This creates a strange distribution of `min i 256`. Better to do `i <- chooseInt (0, min n 256)`.

ChickenProp: Yeah, I wasn't sure about that. I went with this on grounds "larger size means you still get bigger things here on average", but happy to change.

ChickenProp: Done.

MaximilianAlgehed: The problem with the old code is that larger size => n is almost always bigger than 256 => the result is almost always the same.

ChickenProp: Oh yeah, that makes sense.

PR #426, opened 2025-05-15
add withProgress to make testing frameworks get out of the internals

Rewbert:
In an effort to chuck some of the testing frameworks (Tasty, Hspec) out of the internals, I add `withProgress`. Their usage of the internals seemed to be quite innocent (primarily inspecting the testing state to report the current progress by themselves). Here's a combinator that gives an alternative way of doing this.

E.g. Tasty currently does this
```Haskell
import qualified Test.QuickCheck as QC
import qualified Test.QuickCheck.Property as QCP
import qualified Test.QuickCheck.State as QC

quickCheck :: (Progress -> IO ())
           -> QC.Args
           -> QC.Property
           -> IO QC.Result
quickCheck yieldProgress args
  = (.) (QC.quickCheckWithResult args)
  $ QCP.callback
  $ QCP.PostTest QCP.NotCounterexample
  $ \st@QC.MkState {QC.maxSuccessTests, QC.numSuccessTests} _ ->
    yieldProgress $
      if QC.numTotTryShrinks st > 0 then
        emptyProgress {
            progressText = showShrinkCount st
          }
      else
        emptyProgress {
            progressPercent = fromIntegral numSuccessTests / fromIntegral maxSuccessTests
          }
```

and now it is enough to do

```Haskell
import qualified Test.QuickCheck as QC

quickCheck :: (Progress -> IO ())
           -> QC.Args
           -> QC.Property
           -> IO QC.Result
quickCheck yieldProgress args =
  (.) (QC.quickCheckWithResult args)
  $ QC.withProgress $ \p ->
      yieldProgress $
        if QC.numtotalshrinks p > 0 then
          emptyProgress {
              progressText = showShrinkCount p
            }
        else
          emptyProgress {
              progressPercent = fromIntegral (QC.numpassed p) / fromIntegral (QC.maxtests p)
            }
```

Here's the [Hspec](https://github.com/hspec/hspec/commit/ac5b934db4af9b62c48bf3c6a81ec5751188672d) version using this combinator.
Here's the [Tasty](https://github.com/UnkindPartition/tasty/commit/0f52efbd26427b54053119c5acd3ffd6a73f0ce6) version using this combinator.

It makes sense to sit down and figure out a good API for frameworks for interacting with QC, but it takes a while. This is a quick approach to separating testing frameworks from internals. There were some other usages of internals that will need similar treatment (e.g. a framework is importing `Test.QuickCheck.Random` to do something innocent, so there should be a similarly simple fix there).

I may make further PRs in the same spirit as this one. This is in preparation for proposing big changes to the testing loop.

### DIFF
--- a/src/Test/QuickCheck/Property.hs
+++ b/src/Test/QuickCheck/Property.hs
@@ -18,7 +18,7 @@ import Test.QuickCheck.Gen.Unsafe
 import Test.QuickCheck.Arbitrary
 import Test.QuickCheck.Text( isOneLine, putLine )
 import Test.QuickCheck.Exception
-import Test.QuickCheck.State( State(terminal), Confidence(..) )
+import Test.QuickCheck.State( State(terminal, numSuccessTests, numDiscardedTests, maxSuccessTests, numSuccessShrinks, numTryShrinks, numTotTryShrinks), Confidence(..), TestProgress(..) )
 
 #ifndef NO_TIMEOUT
 import System.Timeout(timeout)
@@ -449,6 +449,22 @@ whenFail' m =
     if ok res == Just False
       then m
       else return ()
+
+-- | Performs an IO action every time a property is tested, after every test.
+-- The IO action is allowed to depend on @TestProgress@, which contains information
+-- regarding how testing is progressing.
+withProgress :: Testable prop => (TestProgress -> IO ()) -> prop -> Property
+withProgress m =
+  callback $ PostTest NotCounterexample $ \st _r ->
+    let tp = TestProgress { numpassed    = numSuccessTests st
+                          , numdiscarded = numDiscardedTests st
+                          , maxtests     = maxSuccessTests st
+                          
+                          , numshrinks       = numSuccessShrinks st
+                          , numfailedshrinks = numTryShrinks st
+                          , numtotalshrinks  = numTotTryShrinks st
+                          }
+    in m tp
 
 -- | Prints out the generated test case every time the property is tested.
 -- Only variables quantified over /inside/ the 'verbose' are printed.
--- a/src/Test/QuickCheck/Property.hs
+++ b/src/Test/QuickCheck/Property.hs
@@ -453,16 +453,20 @@ whenFail' m =
 -- | Performs an IO action every time a property is tested, after every test.
 -- The IO action is allowed to depend on @TestProgress@, which contains information
 -- regarding how testing is progressing.
+--
+-- Note: QC invokes callbacks before the internal state has been updated to reflect the
+-- most recent test. The means that e.g. @currentPassed@ will, after the first test has
+-- been executed, still show 0.
 withProgress :: Testable prop => (TestProgress -> IO ()) -> prop -> Property
 withProgress m =
   callback $ PostTest NotCounterexample $ \st _r ->
-    let tp = TestProgress { numpassed    = numSuccessTests st
-                          , numdiscarded = numDiscardedTests st
-                          , maxtests     = maxSuccessTests st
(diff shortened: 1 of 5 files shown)


src/Test/QuickCheck/Property.hs
```diff
@@ -450,6 +450,22 @@ whenFail' m =
...
+-- | Performs an IO action every time a property is tested, after every test.
+-- The IO action is allowed to depend on @TestProgress@, which contains information
+-- regarding how testing is progressing.
+withProgress :: Testable prop => (TestProgress -> IO ()) -> prop -> Property
+withProgress m =
+  callback $ PostTest NotCounterexample $ \st _r ->
+    let tp = TestProgress { numpassed    = numSuccessTests st
+                          , numdiscarded = numDiscardedTests st
+                          , maxtests     = maxSuccessTests st
+
```
MaximilianAlgehed, line 462:
```suggestion
```

src/Test/QuickCheck/State.hs
```diff
@@ -94,5 +94,20 @@ data Confidence =
...
+data TestProgress
+  = TestProgress
+  {
+    numpassed        :: Int -- ^ Number of tests passed so far
+  , numdiscarded     :: Int -- ^ Number of discared tests so far
+  , maxtests         :: Int -- ^ Number of tests to execute
+
+  , numshrinks       :: Int -- ^ Number of successful shrinking steps
+  , numfailedshrinks :: Int -- ^ Number of failed shrinking steps since last successful one
+  , numtotalshrinks  :: Int -- ^ total number of failed shrinking steps
```
MaximilianAlgehed, line 109: Any chance we can have `camelCase` names here?

src/Test/QuickCheck/State.hs
```diff
@@ -94,5 +94,20 @@ data Confidence =
...
+-- | TestProgress, contains information that might be interesting to external
+-- libraries, e.g. Tasty. From this it is possible to install your own callbacks
+-- that reports e.g. progress.
+data TestProgress
+  = TestProgress
+  {
+    numpassed        :: Int -- ^ Number of tests passed so far
+  , numdiscarded     :: Int -- ^ Number of discared tests so far
+  , maxtests         :: Int -- ^ Number of tests to execute
+
```
MaximilianAlgehed, line 106:
```suggestion
```

tests/WithProgress.hs
```diff
@@ -0,0 +1,10 @@
+import Test.QuickCheck
+import Test.QuickCheck.Monadic
+
+main = quickCheck prop_test_withProgress
```
MaximilianAlgehed, line 4: I would really appreciate an actual test here that maybe checks that you get reasonable numbers out of your `TestProgress` datastructure. You could dump the numbers to `IORef`s or similar and fail the test if the numbers we get don't make much sense.

MaximilianAlgehed (changes requested):
I have some minor things here, but I think this is absolutely the right way to go.

src/Test/QuickCheck/State.hs
Rewbert: Absolutely

tests/WithProgress.hs
Rewbert: Sure, that makes more sense. Now I performed the very scientific 'ocular inspection' to assess the correctness!

src/Test/QuickCheck/State.hs
Rewbert: I should also add that these fields are just an axplock of what I thought might be relevant. We can add more fields as needed.

Rewbert:
Think I got them all @MaximilianAlgehed

PR #427, opened 2025-06-10
New instances and CI cleanup

MaximilianAlgehed:


MaximilianAlgehed:
@Rewbert maybe the best thing to do right now is to move the `HackingAgenda.md` file into an issue and trying to make this PR _just_ the instances stuff so that we at least have that on `master`?

Rewbert:
Makes sense. I hope to find some time tomorrow to look at more instances :)

Rewbert:
Pushed some silly instance for `System.Console`. Please see @MaximilianAlgehed 

It exports only three types, and it is trivial to derive instances

```haskell
instance Arbitrary (ArgDescr Int) where
  arbitrary = oneof [ NoArg <$> arbitrary
                    , ReqArg <$> arbitrary <*> arbitrary
                    , OptArg <$> arbitrary <*> arbitrary
                    ]

instance Arbitrary (ArgOrder Int) where
  arbitrary = oneof [ return RequireOrder
                    , return Permute
                    , ReturnInOrder <$> arbitrary
                    ]

instance Arbitrary (OptDescr Int) where
  arbitrary = Option
                <$> arbitrary
                <*> arbitrary
                <*> arbitrary
                <*> arbitrary
```

But it is very unlikely that anything sensible will be generated. However, anything sensitive will be dependent on some context, which we cannot supply lmao. I am not sure that I am a fan of these instances.

Had a look at `System.Posix` as well. It exports a buttload of types. I am sure that we can provide instances for some of them, but not all. Some types are e.g. created via `CType` macros (never used these macros personally before).

MaximilianAlgehed:
CI won't trigger because there are conflicts I think.

Rewbert:
Looks like a silly conflict!

.github/workflows/haskell-ci.yml
```diff
@@ -21,18 +21,15 @@ jobs:
...
                       9.6.6, 9.6.5, 9.6.4, 9.6.3, 9.6.2, 9.6.1,
                       9.4.8, 9.4.7, 9.4.6, 9.4.5, 9.4.4, 9.4.3, 9.4.2, 9.4.1,
                       9.2.8, 9.2.7, 9.2.6, 9.2.5, 9.2.4, 9.2.3, 9.2.2, 9.2.1,
                       9.0.2, 9.0.1,
                       8.10.7, 8.10.6, 8.10.5, 8.10.4, 8.10.3, 8.10.2, 8.10.1,
-                      8.8.4, 8.8.3, 8.8.2, 8.8.1,
-                      8.6.5, 8.6.4, 8.6.3, 8.6.2, 8.6.1,
-                      8.4.4, 8.4.3, 8.4.2, 8.4.1,
-                      8.2.2,
-                      8.0.2]
```
andreasabel, line 35: Why can't you test GHCs older than 8.10?

andreasabel: The more official information (in the .cabal file) is `base >=4.3`, this means you are supporting GHC 7.0 and up.
I think the `tested-with` should be in sync with both CI _and_ the `base` constraint, otherwise there are untested/untried build plans.
So you aim to drop all of GHC < 8.10?

src/Test/QuickCheck/Arbitrary.hs
```diff
@@ -186,6 +196,34 @@ import Data.Functor.Compose
...
 
+#if defined(MIN_VERSION_base)
+#if MIN_VERSION_base(4,16,0)
+import Data.Type.Ord
+#endif
+
+import qualified Data.Semigroup as Semigroup
+import Data.Ord
+
+#if MIN_VERSION_base(4,14,0)
```
andreasabel, line 207: Always true, so you can make this unconditional.

src/Test/QuickCheck/Arbitrary.hs
```diff
@@ -1035,6 +1089,158 @@ instance Arbitrary (f a) => Arbitrary (Monoid.Alt f a) where
   arbitrary = fmap Monoid.Alt arbitrary
   shrink = map Monoid.Alt . shrink . Monoid.getAlt
 #endif
+
+#if MIN_VERSION_base(4,9,0)
```
andreasabel, line 1093: Always true.

src/Test/QuickCheck/Arbitrary.hs
```diff
@@ -1035,6 +1089,158 @@ instance Arbitrary (f a) => Arbitrary (Monoid.Alt f a) where
...
+  arbitrary = mkSolo <$> arbitrary
+  shrink = map mkSolo . shrink . getSolo
+#endif
+
+instance Arbitrary a => Arbitrary (Down a) where
+  arbitrary = fmap Down arbitrary
+  shrink = map Down . shrink . getDown
+#endif
+
+#if MIN_VERSION_base(4,14,0)
```
andreasabel, line 1185: Always true.

andreasabel (commented):
Thanks for the revision.

Looks like some consequences of `base >= 4.14` have not been pushed through yet...

QuickCheck.cabal
```diff
@@ -136,17 +134,6 @@ library
   if !impl(ghc >= 7.4)
```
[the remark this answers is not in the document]
MaximilianAlgehed: technically, `ghc >= 7.4` is false when you're not compiling for `ghc`?

QuickCheck.cabal
```diff
@@ -136,17 +134,6 @@ library
...
-  if impl(ghc >= 7.2) && impl(ghc < 7.6)
-    Build-depends: ghc-prim
-
-  -- Safe Haskell appeared in GHC 7.2, but GHC.Generics isn't safe until 7.4.
-  if impl (ghc < 7.4)
-    cpp-options: -DNO_SAFE_HASKELL
-
   -- random is explicitly Trustworthy since 1.0.1.0
   -- similar constraint for containers
   if impl(ghc >= 7.2)
```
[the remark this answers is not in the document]
andreasabel: The question is whether there is some evidence that QuickCheck works on other compilers than GHC.  If there is no such evidence the question is whether such evidence could be constructed (at all).  
If QuickCheck works outside of GHC, the CI should also deliver that evidence.  Otherwise we are just shooting in the dark.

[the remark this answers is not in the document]
MaximilianAlgehed: It does work on hugs and we do have a CI job for it, so I guess we might as well just try.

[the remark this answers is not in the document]
andreasabel: > the right thing to do would be to change these to `impl(ghc)`.

Yes, this seems to be the conservative way.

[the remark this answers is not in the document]
andreasabel: You could add Hugs to the `tested-with` field then.

src/Test/QuickCheck/Arbitrary.hs
```diff
@@ -186,6 +194,32 @@ import Data.Functor.Compose
...
+import qualified Data.Semigroup as Semigroup
+import Data.Ord
+
+import System.Console.GetOpt
+    ( ArgDescr(..), ArgOrder(..), OptDescr(..) )
+
+import Data.Functor.Contravariant
+
+#if MIN_VERSION_base(4,17,0)
+import Data.Array.Byte
```
Bodigrim, line 211: There is a compatibility package [`data-array-byte`](https://hackage.haskell.org/package/data-array-byte), it would be better to use it when `base < 4.17`, so that `QuickCheck` provides `instance Arbitrary ByteArray` unconditionally.

phadej: > In the interest of maintainability I'm not keen at all to go down the road of compatibility packages. 

The argument is moot, you just shift burden elsewhere.

 `quickcheck-instances`. Copying code: the instance itself and CPP conditionals is not great.

Currently as extra instances exist only in `quickcheck-instances`, it's itself (mostly) CPP free. That is great.

---

@Bodigrim 

> Now quickcheck-instances would have to provide instance Arbitrary ByteArray, but only when built against old versions of base but new versions of QuickCheck. Such conditionals quickly get out of hand.

That is bad, but luckily I'll only support the latest `QuickCheck` in latest `quickcheck-instances`. Supporting older `QuickCheck` versions would indeed be maintenance nightmare.

---

I'm not a fan. I'd prefer `QuickCheck` to use compatibility packages if they define instances for types not-yet in all supported `base`s.

It's easier to users, it's easier to me as `quickcheck-instances` maintainer.  

If users need `Arbitrary ByteArray` it's very likely today that they need that instances also on `GHC <9.4`, so they would still need to depend on `quickcheck-instances` and add an extra import.

TL;DR, adding the instances *conditionally* isn't as helpful as you think.

MaximilianAlgehed: > Please don't downplay their contributions

I'm not downplaying anyone's work I'm talking about a distinction between official `base` and inofficial packages. If you're going to intentionally misinterpreted me we will have no constructive interactions.

phadej: https://hackage.haskell.org/package/data-array-byte

14 direct dependencies including  bytestring, hashable, mempack, primitive, random and text.

I can speak only about `hashable`, but it contains test code which looks like

```
pHash :: [Word8] -> Bool
pHash xs = unsafePerformIO $ withArray xs $ \ p ->
    (hashByteArray (fromList xs) 0 len ==) `fmap` hashPtr p len
```

It would be great to be able to use `ByteArray` instance, but even with change in this PR i wouldn't be able if I want to support GHC <9.4. `hashable` cannot depend on `quickcheck-instances` due dependency loop. I need `ByteArray` instance to exist uniformly across all GHCs, or it's useless to me.

"Luckily" latest hashable is requiring GHC >=9.6, but that is for completely separate reasons, and I would preferred not to have so tight/recent lower bound at this point of time. (The requirement is due changes in `base` which cannot be shimmed, `ByteArray` can).

phadej: > between official base and inofficial packages

TBH, I don't see a real distinction. `base` doesn't have any superb standards. It just happens to be bundled with GHC. As  a concrete example, I don't see any distinction between say `containers` and `primitive`. Latter is not shipped with GHC, because GHC doesnt' depend on it. But I'd argue they are as essential and therefore well maintained packages in the ecosystem as they need to be.

Compatibility packages have/had very respectful maintainers (e.g. `base-compat`). I dare to say they have the highest maintenance standard you'd ever see in Haskell ecosystem. *They wouldn't been used if they hadn't been that good*.

Please, make everyones life easier and add

```cabal
if impl(ghc < 9.4)
  build-depends: data-array-byte
```

I'm very confident that if there's any issue @Bodigrim will resolve it promptly.

Rewbert:
> To clarify, it is an ecosystem-wide convention to use data-array-byte as a compatibility package for ByteArray for GHC < 9.4. For instance, the convention is followed by [bytestring](https://hackage.haskell.org/package/bytestring-0.12.2.0/dependencies), [text](https://hackage.haskell.org/package/text-2.1.2/dependencies), [random](https://hackage.haskell.org/package/random-1.3.1/dependencies), [primitive](https://hackage.haskell.org/package/primitive-0.9.1.0/dependencies), and thus is imposed on thousands of their downstream dependants.

is this the only such circumstance? Making one exception imo isn't too bad. 8.8 is fairly new, even if I think most users are on more modern versions. If we can support some more versions without a lot of trouble, it might be worth it. Depends on the amount of trouble, though.

PR #433, opened 2025-06-24
remove Test.QuickCheck.Gen.Class

Rewbert:
Seemingly unused module. Let's leave the PR open for a while and see if anyone complains?


Rewbert:
@treeowl you originally added this module. Is it still needed for something? Are there any arguments for keeping it around? :)

nick8325:
There's some discussion from when it was added here: #108. Is it now superseded by QuickCheck-GenT?

It looks like what happened is that I removed it from the .cabal file before releasing 2.9.2 (so as not to add new modules in a minor release), and then... forgot to re-add it? See 2e84ad3f. Unfortunately, since it's not mentioned in the .cabal file, it's not exported and I don't think it's possible for anyone to import it.

sol (approved):
LGTM 👍

Rewbert:
It does indeed seem as if a `GenT` would supersede this module.

PR #434, opened 2025-09-28
Make it compile with MicroHs

augustss:
Various minor changes for MicroHs


MaximilianAlgehed:
Thank you Lennart. Do you think you can add CI for microhs too?

MaximilianAlgehed:
We very proudly have CI for hugs! Adding more CI for more implementations is a good thing. Even if it sometimes comes at the cost of a more complex code-base my personal opinion is that Haskell could and should be more than just "what GHC does".

MaximilianAlgehed (approved):

PR #447, opened 2025-10-14
Run CI with lots of stack LTSs to avoid issues with old dependencies etc

### LINKED ISSUE #445: Tests with old libraries
To avoid the bounds issues we had in 2.17 @lehins suggested we could just test against a bunch of stack LTSs. I think that's a good idea that we should be able to set up in CI.

MaximilianAlgehed:
closes #445


Bodigrim:
An easy way to test against old dependencies is `cabal build --prefer-oldest`: it flips the preference of Cabal solver and forces it to choose as old releases as possible, exercising your lower bounds.

PR #452, opened 2025-11-13
Cache dependencies in mhs

MaximilianAlgehed:
This bring the mhs job down from a bit over 5 mins to a bit under 4, the majority of what's left is building mhs itself.


augustss:
I'll take a look.

PR #466, opened 2026-02-12
Fix bug in heap example and silently failing tests

### LINKED ISSUE #465: ToSortedList in heap example fails
```
=== prop_ToSortedList from examples/Heap.hs:121 ===
*** Failed! Falsified (after 5 tests):  
Node (-4) (Node (-3) (Node 1 Empty Empty) (Node 3 Empty Empty)) (Node 4 (Node 4 Empty Empty) Empty)
```

Whether this is intentional, probably not.

(side comment: `main = $quickCheckAll` feels to be a lie, as the test suite doesn't fail if written so)

UlfNorell:
Fixes #465

`$quickCheckAll` has type `IO Bool`, which is allowed as the type of `main`, but a `False` result will not translate to a non-zero exit code. Thus, to use it in a test suite you need something like this
```haskell
main = do True <- $quickCheckAll; return ()
```

### DIFF
--- a/examples/Heap.hs
+++ b/examples/Heap.hs
@@ -144,7 +144,7 @@ instance (Ord a, Arbitrary a) => Arbitrary (Heap a) where
 -- main
 
 return []
-main = $quickCheckAll
+main = do True <- $quickCheckAll; return ()
 
 --------------------------------------------------------------------------
 -- the end.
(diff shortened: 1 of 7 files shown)


MaximilianAlgehed (approved):

MaximilianAlgehed:
An afterthought - maybe `toSortedList` was buggy on purpose?

